Issue #2, opened 2020-02-13
active branch discussion


jodygarnett:
The initial schema plugin is arriving as https://github.com/metadata101/iso19139.ca.HNAP/pull/1 against branch 3.8.x.

We should decide which active branches to maintain:

* **3.6.x** is used by Federal Geospatial Portal (FGP)
* 3.6.x-ec used by Environment Climate Change Canada (ECCC)
* 3.8.x not in use
* **3.10.x** stable branch of core-geonetwork (GeoCat Live, DFO)
* **master** not in use, available to track core-geonetwork master

jodygarnett:
@josegar74 how should we proceed, I think we already know of some small changes needed for 3.10.x compatibility.

jodygarnett:
Going to recommend:

1) start with 3.6.x branch for FGP
2) leave the 3.8.x branch empty (@ianwallen can provide direction here)
3) @josegar74 can work on a 3.10.x branch

ianwallen:
The code that you are currently adding may only work with the 3.8 branch.  I don't think FGP is ready so  submit their changes or if there are plans for them to submit the code as it is specific to there application (may want to verify with them).    So I would recommend the 3.8 branch be the active branch.

jodygarnett:
We have some coordination with FGP, the larger goal here is to combine forces on a common activity.

ianwallen:
We are about to start switching to the 3.10 release (probably after the 3.10.2 release).
So we can probably change the default branch so that it is no longer 3.8

Just wondering, if we want to follow the geonetwork core branch model.  We could place all the 3.10 code into the master and commit to it with cherry picks to the 3.10 release?

ianwallen:
I would also like to propose that we use Travis builds to verify that the commits are not breaking the integration into the geonetwork.   Theoretically, there should be little impact on the plugin build within geonetwork core.- however since this repo was created, I don't believe there has been any successful build with all unit test.  I found one issue related to the autodetect that was causing issues and I have a pull request to fix this.  However after these changes, there are still some errors in the unit test once the plugin has been merged with geonetwork core.  Once these issues are fixed it would be nice to know that similar issues don't come up again.

jodygarnett:
Due to the design of geonetwork we have found issues that need to be resolved in geonetwork core and then back ported to 3.10.x for the plugin to work. Perhaps we can do a better job cross checking the various pull requests from HNAP issues.

[...]

PR #11, opened 2020-03-16
Setup 3.10.x starting from a copy of branch 3.8.x

davidblasby:
Copying from 3.8.x for starting work with GeoNetwork 3.10 for this schema


jodygarnett:
I have updated https://github.com/metadata101/iso19139.ca.HNAP/issues/2 to reflect this change

PR #13, opened 2020-03-19
Auto move thesuruses to data dir

davidblasby:
1) JAR: move schema's thesauruses to resources in jar (see `pom.xml`)
2) SPRING: config-spring-geonetwork.xml modified to setup `SchemaInitializer`
3) JAVA: new `SchemaInitializer` class
    * when the GN datadir is setup, this will be notified (`onApplicationEvent`)
   * it will see if the JAR's thesauruses are in the data dir, if not, it will copy them in

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -49,6 +49,19 @@
     </dependency>
   </dependencies>
   <build>
+    <resources>
+      <resource>
+        <directory>src/main/config/codelist/local/thesauri/theme</directory>
+        <targetPath>thesarus/theme</targetPath>
+      </resource>
+      <resource>
+        <directory>src/main/config/codelist/local/thesauri/place</directory>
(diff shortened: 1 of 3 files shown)


davidblasby:
NOTES: 

* there are two `JeevesApplicationContext`'s created in `JeevesContextLoaderListener`. A parent and child. The `ThesaurusManager` is in the parent, while almost everything else is in the child. This makes pass events from the manager to other classes very difficult. I abandoned this approach.

* The `ThesaurusManager`, on init, sets up a thread to read from the dirs. This waits 10 seconds. There's no message to say it's finished.

* Although there is a little bit of work being done in the handler, it's only ever done once so it shouldn't be an issue.

josegar74 (approved):

PR #17, opened 2020-03-24
Fix bug with initializing schema on Windows systems.

ianwallen:
Changed File.separator to "/" because all url use "/". not os dependent.


davidblasby:
Thanks, @ianwallen for catching this!

PR #23, opened 2020-03-26
make language JSON correct

davidblasby:
I noticed that the data-gn-multilingual-field attribute had "weird" values.  For example;
```json
{
   "eng":"#eng; CAN",
    "fre":"#fra",
    "eng":"#eng"
}
```
This is incorrect, it should be:
```json
{
   "eng":"#eng",
   "fre":"#fra"
}
```
This PR fixes this;

a) update java code so it converts, for document default language, `"eng":"#eng; CAN"` to `"eng":"#eng"`
b) The above creates `{"eng":"#eng","fre":"#fra","eng":"#eng"}` so I added a `distinct-values()` to remove the duplicated value.

### DIFF
--- a/src/main/java/ca/gc/schema/iso19139hnap/util/XslUtilHnap.java
+++ b/src/main/java/ca/gc/schema/iso19139hnap/util/XslUtilHnap.java
@@ -238,7 +238,7 @@ public class XslUtilHnap {
     public static
     @Nonnull String getMappLangId(String lang, String id) {
         if (id == null || id.isEmpty()) {
-            return lang;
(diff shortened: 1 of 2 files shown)


ianwallen:
Although this does fix the issue. Maybe we should fix the reason why get-iso19139.ca.HNAP-other-languages-as-json is getting these values in the first place. It should not be returning duplicates languages and we already have another function for extracting the language from "eng; CAN"

I was hoping to remove getMappedLang as we don't really need 2 ways to get the same results.

josegar74 (approved):

PR #24, opened 2020-03-27
fix saving not working

davidblasby:
This code was doing incorrect things to the output - there were several issues.  The easiest to reproduce would be to take an empty field and just set the french (if eng is the doc default) value.  The change would get "undone".

This fixes all the issues.


ianwallen (approved):
Verified
Thank you for the fix.

Issue #16, opened 2020-03-24
French title not showing up correctly


ianwallen:
If I upload the following file and go to the viewer (taken from FGP/Open Data) and go to the viewer

The English Title looks correct but the French title is incorrect.  It also seems like it is a json array with duplicate values.

["Cartographie des débarquements de la pêche côtière du homard et de l'effort de pêche sur un quadrillage statistique de la région des Maritimes (2012 à 2014)","Cartographie des débarquements de la pêche côtière du homard et de l'effort de pêche sur un quadrillage statistique de la région des Maritimes (2012 à 2014)"]

I have attempted to debug this with no luck.  

I included a sample of the xml file that was uploaded.

[64f741d7-1129-49dd-9e5c-2b1de79024f0.zip](https://github.com/metadata101/iso19139.ca.HNAP/files/4377233/64f741d7-1129-49dd-9e5c-2b1de79024f0.zip)

The record also exist on open data.
https://open.canada.ca/data/en/dataset/64f741d7-1129-49dd-9e5c-2b1de79024f0

This is affecting the 3.8 and 3.10 branch - did not test 3.6

ianwallen:
Found bug - it is related to textgroup title being indexed twice.  I will work on a fix shortly.

josegar74:
I have tried also the iso19139 approach (as described in the PR #26), but indeed something else seem missing as doesn't solve the problem. 

Merged the PR as looks working fine, to investigate further.

Let me know if you'll cherry-pick to 3.10.x or want me to do it, then we can close the issue.

ianwallen:
I have done a cherry-pick of the commit to 3.10.x branch.
Closing the issue.

PR #26, opened 2020-03-30
Fix French title showing up as json array

ianwallen:
Fix for issue #16

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/index-fields/language-default.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/index-fields/language-default.xsl
@@ -64,12 +64,6 @@
             <!-- not tokenized title for sorting -->
             <Field name="_defaultTitle" string="{string($_defaultTitle)}" store="true" index="true"/>
 
-            <xsl:variable name="title"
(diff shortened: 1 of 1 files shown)


josegar74:
Code changes look fine, but I guess can be unified with latest code in `iso19139`. See:

- https://github.com/geonetwork/core-geonetwork/blob/fa2ced925bbea8fc08bbd8cf29053f4c608d6c05/schemas/iso19139/src/main/plugin/iso19139/index-fields/language-default.xsl#L107-L110

- https://github.com/geonetwork/core-geonetwork/blob/fa2ced925bbea8fc08bbd8cf29053f4c608d6c05/schemas/iso19139/src/main/plugin/iso19139/index-fields/language-default.xsl#L114-L122

ianwallen:
I agree that it would be better to incorporate the changes from iso19139

I tried to implement the code from iso19139 but it did not seem to be fixing the issue - so I suspect that there are more changes that need to be added than the lines that you suggested.

Maybe we should have a task to incorporate a lot of the changes from this file as it may fix a lot of the other issues however I would not be the best to do that because I don't know about all the changes that were made for HNAP.

Issue #12, opened 2020-03-16
Clarify HNAP interaction of main language and multiple PT_LOCALE  translations


jodygarnett:
The following has been provided by NrCAN FGP in discussion with @ianwallen:

> The problem is that our geonetwork base is 3.8 and FGP geonetwork base is 3.6.  In geonetwork 3.8 they applied a change to the language setting. 
> So prior to 3.8 if the languages were English(main) and French then they would have a English as the main language and French would be in the PT_LOCALE. So in this case there was only one PT_locale.
> In 3.8, they still have English as the main language but they also have 2 PT_LOCALE – one for each language.  The multiple PT_Locale is causing the issue.
> I have included 2 version of the metadata that were attempted to be loaded.  The 3.6 version loads correctly.  The 3.8 version causes the issues – you test loading them if you want.
> 
> So we either need to apply a transformation prior to uploading the data to FGP so that it remove the main language PT_locale.  Or the FGP system is updated to support multiple PT_LOCALE. I believe it may be easier for us to apply a transformation prior to submitting the data to FGP.
> 
> I have already applied a bunch of fixes related to this issue in the new metadata 101 for HNAP (https://github.com/metadata101/iso19139.ca.HNAP)  for the 3.8 branch - so I presume that someday FGP will be upgraded to support this.

jodygarnett:
The specification has the following example:

> Locale (PT_Locale)
>
> Guideline: NAP registered code lists based on ISO 639.2 alpha-3 terminology codes for language (eng=English, fra=French) and ISO 3166-1 alpha-3 codes (CAN=Canada) for countries shall be used to describe locale and character encoding shall be set to "utf8; utf8", as shown in the following example:
> 
>
> ```xml
>   <gmd:locale>
>     <gmd:PT_Locale id="fra">
>       <gmd:languageCode>
>         <gmd:LanguageCode codeListValue="fra" > codeList="http://nap.geogratis.gc.ca/metadata/register/napMetadataRegister.xml#IC_116">French; Français</gmd:LanguageCode>
>       </gmd:languageCode>
>       <gmd:country>
>         <gmd:Country codeListValue="CAN" codeList="http://nap.geogratis.gc.ca/metadata/register/napMetadataRegister.xml#IC_117" >Canada; Canada</gmd:Country>
>       </gmd:country>
>       <gmd:characterEncoding>
[...]

PR #30, opened 2020-04-02
not put localized character string for default language (redundant)

davidblasby:
This will make it so that the document default language is not put in the alternative language sections.

```:xml
 <gmd:administrativeArea xsi:type="gmd:PT_FreeText_PropertyType">
    <gco:CharacterString>British Columbia</gco:CharacterString>
    <gmd:PT_FreeText>
        **<gmd:textGroup>
            <gmd:LocalisedCharacterString locale="#eng">British
                Columbia</gmd:LocalisedCharacterString>
        </gmd:textGroup>**
        <gmd:textGroup>
            <gmd:LocalisedCharacterString locale="#fra"
                >Colombie-Britannique</gmd:LocalisedCharacterString>
        </gmd:textGroup>
    </gmd:PT_FreeText>
</gmd:administrativeArea>
```

In this example, "British Columbia" is redundant - its in the default section ("CharacterString") and in the alternative section ("#eng").  This will remove the ("#eng") section (in bold).

I also noticed that it was referencing a thesaurus that was in "/local" - however, this is in the "external".

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl
@@ -23,7 +23,7 @@
   <xsl:include href="layout/utility-fn.xsl"/>
 
   <xsl:variable name="thesauriDir" select="XslUtilHnap:getThesauriDir()" />
-  <xsl:variable name="ecCoreThesaurus" select="document(concat('file:///', replace(concat($thesauriDir, '/local/thesauri/theme/EC_Core_Subject.rdf'), '\\', '/')))" />
+  <xsl:variable name="ecCoreThesaurus" select="document(concat('file:///', replace(concat($thesauriDir, '/external/thesauri/theme/EC_Core_Subject.rdf'), '\\', '/')))" />
 
   <xsl:variable name="coreThesaurusEng" select="'Government of Canada Core Subject Thesaurus'" />
   <xsl:variable name="coreThesaurusFre" select="'Thésaurus des sujets de base du gouvernement du Canada'" />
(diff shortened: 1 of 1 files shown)


jodygarnett:
This addresses https://github.com/metadata101/iso19139.ca.HNAP/issues/12

ianwallen:
Does this mean that [#2032](https://github.com/geonetwork/core-geonetwork/pull/2032) is no longer valid?  This had caused lots of issues when upgrading the 3.6 plugin to 3.8 to work in GN 3.8.1.
I thought there was now cases in GN where it was only using the PT_FreeText version once it detected it was a multilingual document?
I will pull down your changes and tested it in our environment to see if I can find some of the old issues that we were having. I will try to provide results before end of day.
If it is no longer an issue then I agree that is would be good to remove the duplication.

davidblasby:
Hi, @ianwallen,

Thanks for looking into this so quickly.

The way the schema works right now is that the document is stores in the non-redundant version (that's what this PR does).

However, it gets expanded to the redundant version when it's sent to the editor.
I think this makes it the best of both worlds.

jodygarnett:
Chating with @josegar74 he would like to explore with core-geonetwork turning off https://github.com/geonetwork/core-geonetwork/pull/2032

Perhaps as a configuration option, or under control of the schema plugin.

ianwallen:
I have tested the changes in our env which uses 3.8 at the moment and it seems to work fine. 

So if @josegar74 is OK with these changes then I don't have any objections with applying these changes.

The only issue that I found was related to the update-fixed-info.xsl is not initially executed for the sample metatadata.  So after I add the sample metadata, I could not edit the english text.  I had to "Save and Close" which executed the update-fixed-info.xsl.  Then I was able to edit the metadata and it worked ok after that point. this is related to [issue 4199](https://github.com/geonetwork/core-geonetwork/issues/4199)

ianwallen:
Just thinking about it again - since this changes removes the duplicate language content it means that the sample metadata data does not really changes when the  update-fixed-info.xsl is executed.  So I'm not really understanding why the editor works correctly after the  update-fixed-info.xsl is executed (save and close).  I have compared the metadata records before and after the  update-fixed-info.xsl and they are pretty much the same with the exceptions of the formatting and some minor data corrections.

I guest what I'm not understanding is your statement.

> "However, it gets expanded to the redundant version when it's sent to the editor."

But this would not be the first time I don't understand something in geonetwork :)

Either way, this issue should be investigated - maybe as another issue.

josegar74:
@ianwallen for me the samples can be edited fine. I have checkout latest Geonetwork 3.10.x branch and the 3.10.x of this schema.

josegar74:
In any case, going to merge it. I've done some additional tests and looks fine.

davidblasby:
@ianwallen  - I will take a look and see if the samples are working.  I'm pretty sure they are.

However, some of my previous PRs made some fixes for issues like what you're describing - like #24

davidblasby:
I can verify 3.10.x GN is working fine for modifying the sample data (first load).

ianwallen:
Thank you for confirming.  It is possible that the version I tested which is based on 3.8 does not have all the latest fixes.  We are planning to upgrade to 3.10 next week.  If I see this issue again after our upgrade then I will raise it as a new issues.

PR #32, opened 2020-04-09
use new GN-core multi-entry-combiner for organization name

davidblasby:
The diff is a bit more complicated because I edited some already commented out code (from the EC).

The only changes in layout-custom-fields is the addition of this template - <xsl:template mode="mode-iso19139" match="gmd:organisationName" priority="2000">

I made some very minor changes to config-edit because if you don't mark something as multi-lingual, the multilingual sections will be removed by update-fixed-info!

This just converts the gmd:organizationName into a custom HTML fragment.  There is a new multi-entry-combiner directive in GN-core that this directly uses.

There is a full example (config and html fragment) that this produces in MultiEntryCombiner.js - see PR, below).

You need these GN-core PRs (backported to 3.10.x) to use this;
https://github.com/geonetwork/core-geonetwork/pull/4597 (contains the directive)
https://github.com/geonetwork/core-geonetwork/pull/4596 (fixes to the keyword picker so it can be used by the multi-entry-combiner directive)

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
@@ -194,12 +194,12 @@
           },
 
           {
-            "type": "thesurus",
+            "type": "thesaurus",
             "heading": {
               "eng": "Government of Canada Organization",
               "fra": "Organisation du Gouvernement du Canada"
             },
-            "thesurus": "external.theme.EC_Government_Titles"
+            "thesaurus": "external.theme.EC_Government_Titles"
           },
 
           {
--- a/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
@@ -94,177 +94,149 @@
     </xsl:call-template>
   </xsl:template>
 
-<!--
 
+
+
+<!--
+    Special handling for the gmd:organisationName.  See MultiEntryCombiner.js for more info on how this works.
+    Basically this is a replacement for render-element.
+    Instead, it just creates a VERY simple HTML fragment (see MultiEntryCombiner.js for example).
+    The MultiEntryCombiner directive handles most of this.
+    This code mostly sets up two things;
+    a) basic shell HTML so it can be displayed in the editor (see MultiEntryCombiner.js for HTML example).
+    b) sets up the JSON configuration (see MultiEntryCombiner.js for example JSON).
+-->
   <xsl:template mode="mode-iso19139" match="gmd:organisationName" priority="2000">
 
     <xsl:variable name="xpath" select="gn-fn-metadata:getXPath(.)"/>
     <xsl:variable name="isoType" select="if (../@gco:isoType) then ../@gco:isoType else ''"/>
     <xsl:variable name="labelConfig" select="gn-fn-metadata:getLabel($schema, name(), $labels, name(..), $isoType, $xpath)"/>
 
-   <xsl:variable name="theElement" select="." />
-
-    <xsl:variable name="values">
-      <values>
-        &lt;!&ndash; Or the PT_FreeText element matching the main language &ndash;&gt;
-        <xsl:if test="gco:CharacterString">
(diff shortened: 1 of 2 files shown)


davidblasby:
@josegar74 - please review.

DO NOT MERGE until the 2 GN-core PRs are merged and backported to GN 3.10.x!

ianwallen:
Sent more detail review to David.  But the main issue I see at the moment is that this change does not seem to allow organizations other than Government of Canada.  I believe it needs to be modified to allow other 3rd party organizations to be entered.

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
```diff
@@ -94,177 +94,149 @@
...
+                "fra": ""
+            },
+            "values": {
+              "eng": "Government of Canada",
+              "fra": "Gouvernement du Canada"
+            }
+          },
+
+          {
+            "type": "thesurus",
```
ianwallen: Typo - I think this should be "thesaurus"

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
```diff
@@ -94,177 +94,149 @@
...
+            "heading": {
+              "eng": "Government of Canada Organization",
```
ianwallen, line 199: Not sure why the word "Organization" is in the name?

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
```diff
@@ -94,177 +94,149 @@
...
+              "fra": "Organisation du Gouvernement du Canada"
+            },
+            "thesurus": "external.theme.EC_Government_Titles"
```
ianwallen: Another Typo - I think this should be "thesaurus"

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
```diff
@@ -94,177 +94,149 @@
...
+    <!--
+      Creates this (actual values of the metadata record);
 
-    <xsl:variable name="xpath" select="gn-fn-metadata:getXPath(.)"/>
-    <xsl:variable name="isoType" select="if (../@gco:isoType) then ../@gco:isoType else ''"/>
-    <xsl:variable name="labelConfig" select="gn-fn-metadata:getLabel($schema, name(), $labels, name(..), $isoType, $xpath)"/>
-
-    <xsl:variable name="theElement" select="." />
+        <values>
+            <value ref="15" lang="eng">Government of Canada; Library of Parliament; david place2</value>
```
ianwallen: "david place2" not the best example :)
Here are some [examples from Public Services and Procurement Canada](https://www.tpsgc-pwgsc.gc.ca/apropos-about/rgnstnnll-rgnztnal-eng.html)

ianwallen (changes requested):
Found a few typo's and small fixes.

The main issue is that it does not seem to allow for other organizations to be added.

davidblasby:
> Sent more detail review to David. But the main issue I see at the moment is that this change does not seem to allow organizations other than Government of Canada. I believe it needs to be modified to allow other 3rd party organizations to be entered.

Hi, Ian,

If as I said in the email, if we have a thesaurus with a list of organizations, i can very easily change the "fixedValue" ("G.Of.C") to a thesaurus keyword picker.

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
davidblasby: This is a title (heading) that's put on the UI to help users.

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
davidblasby: ;)

davidblasby:
@ianwallen - see the changes to GN-core
For here, I've fixed "thesaurus" and the example.
In GN-core, i've got two language choices - the UI lang and the data-entry lang.

davidblasby:
I am still working on this, but there's too many open pull requests on different branches.  
I will consolidate and continue this in another PR.

PR #33, opened 2020-04-09
removed getMappedLang as it was not longer needed since the 3.10.2 co…

ianwallen:
Removed getMappedLang as it was not longer needed for the 3.10.2

Also getMappedLang seems to break an api test in the unit test. I did not identify why it was causing the issue but when removed, it seems to have fixed the issue.

### DIFF
--- a/src/main/java/ca/gc/schema/iso19139hnap/util/XslUtilHnap.java
+++ b/src/main/java/ca/gc/schema/iso19139hnap/util/XslUtilHnap.java
@@ -206,46 +206,6 @@ public class XslUtilHnap {
         }
         return defaultValue;
     }
-
-    /**
-     * These two functions map {"eng; CAN":"#", "fra":"fra"} to '{"eng":"#eng; CAN", "fre" : "#fra"}' in utility-tpl-multilingual.xsl
-     *
-     * @param lang  hnap language code
-     * @return  iso639 language code
-     */
-    public static
-    @Nonnull
-    String getMappedLang(String lang) {
-        switch(lang) {
-            case "eng; CAN":
-                return "eng";
-
-            case "fra":
-            case "fra; CAN":
-                return "fre";
-
-            default:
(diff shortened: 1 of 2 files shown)


josegar74:
@ianwallen I think probably due to the merge first of https://github.com/metadata101/iso19139.ca.HNAP/pull/34 has caused a conflict, please check to update, apologies.

ianwallen:
The conflict have been fixed.

ianwallen:
@davidblasby . This change reverts the changes that you had previously done to get-iso19139.ca.HNAP-other-languages-as-json.  The problem what that it was not getting the language correctly in the first place. Now that 3.10.2 has better language handling it should no longer get bad language values so there is no need to distinct the values or use the getMappedLang.

I'm not sure if you had a specific testcase that you were using which caused bad values in the language?

davidblasby:
@ianwallen 

The test case is very easy - just look at the configuration sent to the browser.  
I just use the browser's dev tool and look at the "editor?" request.

  data-gn-multilingual-field="{&#34;eng&#34;:&#34;#eng&#34;,&#34;fre&#34;:&#34;#fra&#34;}"

^^ you'll see lots like this

I then just decode it here - https://codebeautify.org/html-decode-string

  data-gn-multilingual-field="{"eng":"#eng","fre":"#fra"}"

Test case -- it should just have "eng" and "fre" in the dictionary.  "eng" should NOT appear twice.

ianwallen:
@davidblasby 
Looks like I had forgotten to apply the switch from fra to fre.
I have made the corrections and I now get the following values which seems to match the expectations of what you had posted.
 
data-gn-multilingual-field="{"eng":"#eng","fre":"#fra"}"

davidblasby:
@ianwallen - awesome!

josegar74 (approved):
Code changes look fine, thanks.

PR #35, opened 2020-04-17
improve org name selector (and bug fix)

davidblasby:
Changes to Org name selector;

a) changed the first part ("Government of Canada") to be a keyword picker that defaults to "Government of Canada".  The included thesaurus can be added to.  Users can just type in whatever they want as well.

b) fixed an issue with saving.  Problem was when removing the default language value.  The change wouldn't actually take place.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
@@ -182,14 +182,15 @@
     <xsl:variable name="json_config">
       [
           {
-            "type": "fixedValue",
+            "type": "thesaurus",
             "heading": {
-                "eng": "",
-                "fra": ""
+                "eng": "Main Organization Enitity",
+                "fra": "Main Organization Enitity (fr)"
             },
-            "values": {
-              "eng": "Government of Canada",
-              "fra": "Gouvernement du Canada"
+            "thesaurus": "external.theme.EC_Org_Names",
+            "defaultValues": {
+                  "eng": "Government of Canada",
+                  "fra": "Gouvernement du Canada"
             }
           },
 
--- a/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
@@ -184,8 +184,8 @@
           {
             "type": "thesaurus",
             "heading": {
-                "eng": "Main Organization Enitity",
-                "fra": "Main Organization Enitity (fr)"
+                "eng": "",
+                "fra": ""
             },
             "thesaurus": "external.theme.EC_Org_Names",
             "defaultValues": {
@@ -197,8 +197,8 @@
           {
             "type": "thesaurus",
             "heading": {
-              "eng": "Government of Canada Organization",
-              "fra": "Organisation du Gouvernement du Canada"
+              "eng": "Department/Agency",
+              "fra": "Département/agence"
             },
             "thesaurus": "external.theme.EC_Government_Titles"
           },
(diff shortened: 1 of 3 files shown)


src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
```diff
@@ -182,14 +182,15 @@
     <xsl:variable name="json_config">
       [
           {
-            "type": "fixedValue",
+            "type": "thesaurus",
             "heading": {
-                "eng": "",
-                "fra": ""
+                "eng": "Main Organization Enitity",
```
ianwallen: typo "Entity"?
But I also recommend not having this title.
But if it is kept then translation is required in next line

ianwallen (changes requested):
I send an email regarding a full review - for which I wanted ECCC to also see my comments.

ianwallen:
Based on email discussions, it sounds like the following should be done.

- Remove: "Main Organization Enitity" header and just let the first box go beside the main field tiltle of "Organization Name"

- The second title(currently Government of Canada Organization) should be changed to the following:
EN - “Department/Agency”
FR - “Département/agence”

davidblasby:
Ok, text change.
Also, fixed some layout issues (add commit to GN-core)

ianwallen:
Could you please reference the pull request from GN-core that fixed some layout issues.

davidblasby:
https://github.com/geonetwork/core-geonetwork/pull/4652
(its already been merged to master)

ianwallen:
That pull request is not flagged for 3.10.x.   I don't believe we should have dependencies on master? Maybe you can make a request on that pull request to have it flagged for 3.10.x?

davidblasby:
I was going to make a backport PR.
I don't think I have permission on the PRs to attach to milestones...

davidblasby:
https://github.com/geonetwork/core-geonetwork/pull/4659
^^ backport PR

davidblasby:
@ianwallen - backport is merged.  This should be ok to merge now?
I can still work on it if there are any issue/changes.  
I want to consolidate where we are and get everything merged.

ianwallen:
@davidblasby  It looks good.  I only have 2 issues.

- At one point you mentioned that you had disabled the "all" in the language and I think it should be enabled again to be consistent with the rest of the interface.  

- Also, I'm not able to get these changes to work when I test it using the geonetwork core 3.10.x branch.  I can only get it to work when using the master branch.  So it looks like the 3.10.x branch is missing some cherry picks.   Could you look into this and identify the missing commits and maybe ask someone to cherry pick those commits. 

Thank you

davidblasby:
@ianwallen  - I am trying to get all the commits into 3.10.x.  Its a process...

I am skeptical with adding an "all".  The other "all" buttons are just for a single fields - it's unclear how this should look as its actually 3 fields somewhat combined.

What should the first 2 fields look like when "all" is pressed?

davidblasby:
I currently have 6 open HNAP PRs - I think we need to consolidate before doing any new work.

ianwallen:
Is this organization a sort of block/group input?  Is why it is more complex? I would think that each of the 3 fields would have the language selection (all/one/eng/fra) as I believe each once can be changed independent of the other.  Or they could all change as a group - as they will have the eng/fra in the top right of the text box it would be easy to identify the language of each box.

As this may be a little more complex - we can look at this one as a separate PR in the future.

So if you can track down the missing cherry pick commits and make it so this PR works on the 3.10.x branch then I will approve this PR.

davidblasby:
I'm going to go through all the PR/commits I've made and see which ones aren't in 3.10.x

davidblasby:
We can go through the org UX again and I can fix it up so everyone is happy.

PR #37, opened 2020-04-23
add english-only template (for other canadian projects)

davidblasby:


ianwallen:
No objection from me for this this pull request.

davidblasby:
@ianwallen - Not sure how to email you.

I'm having issues with 3.10.x and HNAP.  The test cases are not running - and the layout of the hnap is looking wrong.
I'm trying to track down the problem - 

I did see this (but I am having troubles with the debugger at the moment):

Error at xsl:value-of on line 90 column 173 of utility-tpl-multilingual.xsl:
  XPST0017: XPath syntax error at char 52 on line 90 in {...pedLang($mainLanguage), '":...}:
    Cannot find a matching 1-argument function named
  {java:ca.gc.schema.iso19139hnap.util.XslUtilHnap}getMappedLang()
Error at xsl:value-of on line 99 column 163 of utility-tpl-multilingual.xsl:
  XPST0017: XPath syntax error at char 53 on line 99 in {...edLang($codelistValue), '":...}:
    Cannot find a matching 1-argument function named
  {java:ca.gc.schema.iso19139hnap.util.XslUtilHnap}getMappedLang()

However, this doesn't actually seem to be the error.

3.10.x -- test cases run fine
3.10.x + HNAP -- test case fail

ianwallen:
I suggest that you try to pull the changes from pull request #33 - I believe it may fix the issue.

davidblasby:
@ianwallen  - actually, the issue was that you cannot "add-schema.sh" the HNAP and then run the GN-core test cases.  The test cases do not depend on the HNAP jar, but bring in the XSLT.  Since the XSLT references functions in the HNAP jar, it gets the error.

Solution (at least for now) is to NOT run test cases in GN-core if you've installed a schema that has its own XSLT java functions.

ianwallen:
@davidblasby 

@josegar74  approved #33 today and I have brought the changes into our pipeline. The HNAP was added using the add-schema.sh and our pipeline builds seems to work correctly now.  You can try to merge the latest changes into your branch to see if you still have the problem.

Our pipeline builds geonetwork with all the tests. It is not really an option for us at this time to disable the unit tests in the pipeline builds because we do make changes to the geonetwork core and we need to run the test to ensure our change do not break some basic functionality.

Hopefully the changes that you have been doing in the java code with the new pull request will not re-introduce this issue.  It is good that you have identified the problem - hopefully we can avoid hitting the issue or possibly do a fix in geonetwork core to avoid the problem.

src/main/plugin/iso19139.ca.HNAP/templates/vector-unilingual.xml
```diff
@@ -0,0 +1,383 @@
...
+  </gmd:language>
+  <gmd:characterSet>
+    <gmd:MD_CharacterSetCode codeList="http://nap.geogratis.gc.ca/metadata/register/napMetadataRegister.xml#IC_95" codeListValue="RI_458">utf8; utf8</gmd:MD_CharacterSetCode>
+  </gmd:characterSet>
+  <gmd:hierarchyLevel>
+    <gmd:MD_ScopeCode codeList="http://nap.geogratis.gc.ca/metadata/register/napMetadataRegister.xml#IC_108" codeListValue="RI_622">dataset; jeuDonnées</gmd:MD_ScopeCode>
+  </gmd:hierarchyLevel>
+  <gmd:contact>
+    <gmd:CI_ResponsibleParty>
+      <gmd:organisationName xsi:type="gmd:PT_FreeText_PropertyType" gco:nilReason="missing">
```
josegar74: The attribute `xsi:type="gmd:PT_FreeText_PropertyType"` should be removed in all the elements using it in this template.

davidblasby: I removed these...

davidblasby:
@josegar74 - please merge

PR #41, opened 2020-05-05
Align isMainLanguageEmpty check with iso19139 checking gco:CharacterString

josegar74:


davidblasby:
I did the same thing (yesterday) and didn't notice any problems.

PR #43, opened 2020-05-05
fix how multilingual elements are being referensed

davidblasby:
@josegar74 - this does was we talked about this morning.
However,
a) I had to do a little more since it was sending BOTH the <charstring> and the corresponding (same lang) _lang_eng_XYZ
b) I have updated the multi-edit-combiner-directive to work in the same manner

This REQUIRES;
 https://github.com/geonetwork/core-geonetwork/pull/4664


davidblasby:
The refs will now be like this;

a) default lang -> will point DIRECTLY to the <characterString> ref
b) alt lang -> will point DIRECTLY to the PT text ref (if it exists) 
c) alt lang -> will point INDIRECTLY to a non-existent PT - _lang_fr_XYZ (if it doesnt exist)
    * XYZ is the PARENT element

josegar74 (approved):

PR #44, opened 2020-05-06
better support for uni-lingual records

davidblasby:
This was originally noticed by @ianwallen and I talked with @josegar74  about it.

I've fixed up a few items with how HNAP deals with uni-lingual elements.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/templates/vector-unilingual.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/templates/vector-unilingual.xml
@@ -16,7 +16,7 @@
     <gco:CharacterString/>
   </gmd:fileIdentifier>
   <gmd:language>
-    <gco:CharacterString>eng</gco:CharacterString>
+    <gco:CharacterString>eng; CAN</gco:CharacterString>
   </gmd:language>
   <gmd:characterSet>
     <gmd:MD_CharacterSetCode codeList="http://nap.geogratis.gc.ca/metadata/register/napMetadataRegister.xml#IC_95" codeListValue="RI_458">utf8; utf8</gmd:MD_CharacterSetCode>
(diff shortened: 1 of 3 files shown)


josegar74 (approved):

src/main/plugin/iso19139.ca.HNAP/templates/vector-unilingual.xml
```diff
@@ -104,7 +120,7 @@
       <gmd:citation>
         <gmd:CI_Citation>
           <gmd:title >
-            <gco:CharacterString>English-only HNAP</gco:CharacterString>
+            <gco:CharacterString>English-only HNAP2</gco:CharacterString>
```
ianwallen, line 123: Why is this called HNAP2?

src/main/plugin/iso19139.ca.HNAP/templates/vector-unilingual.xml
```diff
@@ -82,6 +82,22 @@
   <gmd:dataSetURI gco:nilReason="missing">
     <gco:CharacterString/>
   </gmd:dataSetURI>
+  <gmd:locale>
+    <gmd:PT_Locale id="eng">
+      <gmd:languageCode>
+        <gmd:LanguageCode codeList="http://nap.geogratis.gc.ca/metadata/register/napMetadataRegister.xml#IC_116"
+                          codeListValue="eng">English; Anglais</gmd:LanguageCode>
```
ianwallen, line 89: Do you find it odd that a uni-lingual record has bilingual code values?
This may be a hard one to fix!

ianwallen (commented):
David,
Just noticed a couple odd changes in this pull request.

src/main/plugin/iso19139.ca.HNAP/templates/vector-unilingual.xml
davidblasby: I was trying to verify that changes were being picked up.  The inflate-metadata.xsl was undoing changes that I was making.  I'll change it back in the future.

src/main/plugin/iso19139.ca.HNAP/templates/vector-unilingual.xml
davidblasby: yes - definitely weird.  I think we need to verify we are using the correct code list....

This might be ok -- however, not 100% sure about a uni-lingual HNAP (or would that just be NAP)?

PR #52, opened 2020-06-05
add titles to thesauri and force overwrite during initialization

davidblasby:
There are two changes;

a) added <title> the rdfs.  
    * these will be used in the metadata record instead of "EC_Content_Scope.rdf"
    * these show up in the UI like this;
image
b) I've set the hnap schema initializer to always deploy new schemas (overwrite existing ones).  This should be ok since the external thesauri are considered read-only in GN (and, therefore, cannot be modified by users).

NOTE: looks like the editor deleted some trailing spaces.  Sorry for the fluff in the commit.

### DIFF
--- a/src/main/config/codelist/local/thesauri/theme/EC_Content_Scope.rdf
+++ b/src/main/config/codelist/local/thesauri/theme/EC_Content_Scope.rdf
@@ -2,7 +2,14 @@
 <rdf:RDF
 	xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
 	xmlns:ns2="http://www.w3.org/2004/02/skos/core#"
-	xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#">
+	xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
+	xmlns:skos="http://www.w3.org/2004/02/skos/core#"
+	xmlns:dc="http://purl.org/dc/elements/1.1/">
+
+  <skos:ConceptScheme rdf:about="http://geonetwork-opensource.org/thesaurus/naturalearth-and-seavox">
+    <dc:title>EC Content Scope</dc:title>
+    <dc:description></dc:description>
+  </skos:ConceptScheme>
 
   <rdf:Description rdf:about="http://geonetwork-opensource.org/EC/contentscope#addapplicabledatacontent">
     <ns2:scopeNote xml:lang="en">Definition</ns2:scopeNote>
@@ -31,13 +38,13 @@
     <rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
 	<ns2:prefLabel xml:lang="en">Party (Individual/Organization)</ns2:prefLabel>
   </rdf:Description>
-  
+
   <rdf:Description rdf:about="http://geonetwork-opensource.org/EC/contentscope#facility">
     <ns2:scopeNote xml:lang="fr">Definition</ns2:scopeNote><ns2:scopeNote xml:lang="en">Definition</ns2:scopeNote><ns2:prefLabel xml:lang="fr">Fonctionnalité</ns2:prefLabel>
     <rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
 	<ns2:prefLabel xml:lang="en">Facility</ns2:prefLabel>
   </rdf:Description>
-  
+
   <rdf:Description rdf:about="http://geonetwork-opensource.org/EC/contentscope#substance">
     <ns2:scopeNote xml:lang="fr">Definition</ns2:scopeNote><ns2:scopeNote xml:lang="en">Definition</ns2:scopeNote><ns2:prefLabel xml:lang="fr">Substance</ns2:prefLabel>
     <rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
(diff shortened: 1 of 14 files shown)


davidblasby:
@ianwallen - I'm making some changes to the thesauri (see other commits).

Just wanted to make sure this was called out to you now.

NOTE: I am currently working on removing the "Government of Canada Core Subject Thesaurus/Thésaurus des sujets de base du gouvernement du Canada" hard coded code inside HNAP.  And instead use the thesaurus title (which will likely look something like "Government of Canada Core Subject Thesaurus/Thésaurus des sujets de base du gouvernement du Canada" - since rdf and GN-core do not allow multi-lingual titles).

This will mean the english/french thesaurus name will be the SAME.  
This change should only affect how the core subject is being used (the others were just using the file name).

davidblasby:
@ianwallen - note - I've asked for more official titles and will change them when I get them.

src/main/java/ca/gc/schemas/iso19139hnap/init/SchemaInitializerThesauri.java
```diff
@@ -54,7 +56,9 @@ public void ConfigureThesauruses(Path thesauriDir) throws IOException {
             Files.createDirectories(Paths.get(thesauriDir.toString(), resourceTypeDir, "thesauri", dir));
             //full path to put the .rdf
             Path correspondingDataDirFile = Paths.get(thesauriDir.toString(), resourceTypeDir, "thesauri", dir, fname);
-            if (!Files.exists(correspondingDataDirFile)) {
+            if (!Files.exists(correspondingDataDirFile) || OVERWRITE_EXISTING_THESAURI) {
+                if (Files.exists(correspondingDataDirFile))
+                    Files.delete(correspondingDataDirFile);
```
ianwallen, line 61: Could we update the log messages so that if it is new insert then the log say "copying" and if it is overwriting (delete/copy) it indicates that it is "overwriting".  Or it could log a delete message before the copy message.  I don't have any preference but I believe the logs message should be clarified.

ianwallen (approved):
Looks good -  I don't see any issues except for a small logging issue and the missing translation on the conceptSchema titles which you indicate is not supported.  I'm no expert on the RDF but would the following work for multilingual titles?
https://www.w3.org/wiki/SkosCoreGuideToc/SectionMultilingual 

At some point, I would also like to see a PR that would rename the none ECCC specific thesaurus files to GC. I have created  issue #53 for this.

src/main/config/codelist/local/thesauri/theme/EC_Content_Scope.rdf
```diff
@@ -2,7 +2,14 @@
 <rdf:RDF
 	xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
 	xmlns:ns2="http://www.w3.org/2004/02/skos/core#"
-	xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#">
+	xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
+	xmlns:skos="http://www.w3.org/2004/02/skos/core#"
+	xmlns:dc="http://purl.org/dc/elements/1.1/">
+
+  <skos:ConceptScheme rdf:about="http://geonetwork-opensource.org/thesaurus/naturalearth-and-seavox">
+    <dc:title>EC Content Scope</dc:title>
```
ianwallen, line 10: All others have ECCC but this one only has EC? Is that on purpose?

davidblasby: I expect I will be updating all of the titles with more "official" versions.
I will fix this...

src/main/java/ca/gc/schemas/iso19139hnap/init/SchemaInitializerThesauri.java
davidblasby: Good point - I will add.

davidblasby:
> Looks good - I don't see any issues except for a small logging issue and the missing translation on the conceptSchema titles which you indicate is not supported. I'm no expert on the RDF but would the following work for multilingual titles?
> https://www.w3.org/wiki/SkosCoreGuideToc/SectionMultilingual

This looks like its more for the multi-lingual concepts - which we're already using (fr and eng).  
I'm not sure if you can put multilingual "title"s in the 1st section.  
And, even if you could, the rest of GN wouldn't handle them.

After looking a bit more, I noticed that there's a short metadata section about the thesaurus in the metadata record - i.e. dates and responsible party.
HNAP currently puts in hard-coded values for this - but these will likely be missing when I remove the "core thesaurus" hard coded stuff. 

I'll send you an example next week...

davidblasby:
Actually, here's an example that indicates you can have multi-lingual titles;
https://www.w3.org/wiki/SkosCoreGuideToc/SectionVersioning

However, this is isn't really supported in GN nor in how the schemas work.
However, it is something that we could add in update-fixed-info -- but that will likely cause issues in some other processing chains.

I will look more into it monday.

jodygarnett:
Is this PR still active?

geothorne:
Has this functionality been pushed to master?
Lee and I have yet to actually cleaning up and add titles to the ECCC and
GC custom (outside of ca HNAP) thesauri

davidblasby:
@jodygarnett @geothorne  - this functionality was already merged via another PR

PR #56, opened 2020-06-16
Multilingual titles

davidblasby:
This allows multilingual thesaurus titles.

I've also removed the core-thesaurus special-case support (which is redundant now).

Theres also some bug fixes (thesaurus).

SEE https://github.com/geonetwork/core-geonetwork/pull/4766


davidblasby:
@josegar74

davidblasby:
@ianwallen  - I noticed that HNAP was "overwriting" iso19139 templates.  That meant that if you install hnap, it will change is19139 - causing issues.  I've made explicit schema check to prevent this.  I noticed this is both the editor and how the thesaurus transformations were being done.  

@josegar74  - please review.  take extra look at layout and layout-custom-fields.

Also, another commit was added to gn-core -- https://github.com/geonetwork/core-geonetwork/pull/4766

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
```diff
@@ -84,7 +87,7 @@
   </xsl:template>
 
   <!-- Readonly elements -->
-  <xsl:template mode="mode-iso19139" priority="2005" match="gmd:fileIdentifier|gmd:dateStamp">
+  <xsl:template mode="mode-iso19139" priority="2005" match="gmd:fileIdentifier[$schema = 'iso19139.ca.HNAP']|gmd:dateStamp[$schema = 'iso19139.ca.HNAP']">
```
josegar74: This template can be removed, as can rely in the `iso19139`  template to handle these elements. Tested working fine.

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
```diff
@@ -256,7 +259,7 @@
 
 
   <!-- Distribution format: Show list of allowed formats -->
-  <xsl:template mode="mode-iso19139" match="//gmd:distributionInfo/gmd:MD_Distribution/gmd:distributionFormat/gmd:MD_Format/gmd:name" priority="2005">
+  <xsl:template mode="mode-iso19139" match="//gmd:distributionInfo/gmd:MD_Distribution/gmd:distributionFormat/gmd:MD_Format/gmd:name[$schema = 'iso19139.ca.HNAP']" priority="2005">
```
josegar74, line 241: To check if can be better to remove this template and use the `data-gn-keyword-picker` in `config-editor.xml`, like:

```
  <for name="gmd:name" xpath="//gmd:distributionInfo/gmd:MD_Distribution/gmd:distributionFormat/gmd:MD_Format/gmd:name"
         use="data-gn-keyword-picker">
      <directiveAttributes data-thesaurus-key="external.theme.EC_Resource_Formats" data-focus-to-input="false"/>

  </for>
```

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
```diff
@@ -293,7 +296,7 @@
     </xsl:call-template>
   </xsl:template>
 
-  <xsl:template mode="mode-iso19139" match="gmd:EX_GeographicBoundingBox" priority="2005">
+  <xsl:template mode="mode-iso19139" match="gmd:EX_GeographicBoundingBox[$schema = 'iso19139.ca.HNAP']" priority="2005">
```
josegar74: This template can be removed, as can rely in the `iso19139` template to handle these elements. Tested working fine.

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
```diff
@@ -353,7 +356,7 @@
 
 
   <xsl:template mode="mode-iso19139" priority="2005"
-                match="gmd:linkage1111">
+                match="gmd:linkage1111[$schema = 'iso19139.ca.HNAP']">
```
josegar74: This template is not really used, can be removed. Seem was added some dummy value in the match to disable it (`gmd:linkage1111`), this elemetn is handled in the online resources panel

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
davidblasby: I will not action this in this PR -- lets move this out later...

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
davidblasby: removed

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
davidblasby: removed

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
davidblasby: removed

src/main/plugin/iso19139.ca.HNAP/convert/thesaurus-transformation.xsl
```diff
@@ -179,12 +240,36 @@
...
     </gmd:MD_Keywords>
   </xsl:template>
 
+  <!-- given a thesarus and language, find the most appropriate title-->
+  <xsl:function name="geonet:getThesaurusTitle">
+    <xsl:param name="thesarus" />
+    <xsl:param name="lang1" />
+
+    <xsl:variable name="lang" select="lower-case($lang1)" />
+    <xsl:variable name="lang_2letter" select="lower-case(XslUtilHnap:twoCharLangCode($lang))" />
```
ianwallen, line 254: @davidblasby 
This line causes the following error in the build when executing the unit tests

Error at xsl:variable on line 254 column 97 of thesaurus-transformation.xsl:
  XPST0017: XPath syntax error at char 45 on line 254 in {...ilHnap:twoCharLangCode($lan...}:
    Cannot find a matching 1-argument function named
  {java:ca.gc.schema.iso19139hnap.util.XslUtilHnap}twoCharLangCode()

It fails in the GetKeywordByIdTest

Could you please correct this.

Thank you

davidblasby: This function exists;
https://github.com/metadata101/iso19139.ca.HNAP/blob/3.10.x/src/main/java/ca/gc/schema/iso19139hnap/util/XslUtilHnap.java#L172

I'm going to guess you are running the GN unit tests when you have HNAP installed.  
You cannot do that - the modules do NOT have dependencies on HNAP (specifically, the jar), but pick up hnap because the schema files (but not the jar) been installed in the web app.  This causes errors.

ianwallen: @davidblasby 
Documentation specifically has instruction where it is acceptable to include the module and do a build (with or without tests) - it is even with the HNAP documentation on the main.[readme.md]https://github.com/metadata101/iso19139.ca.HNAP)

In our case we have added some extra java code that we compile as part of the regular build and this code has unit test which are also executed as part of our build pipeline.  So in our case we need to be able to build and run unit test for the whole system or we cannot be deploy changes.  This is standard practice.

So if the build with these changes generate errors then maybe the unit tests are flawed.  There seems to be a bug  where the unit test are being executed on all the schema plugins instead of only the ones that are in the geonetwork core code.  If it is going to be executed on other plugins then the unit test should ensure that they have all the dependencies fully loaded. Is it possible to fix the unit test?

davidblasby: @ianwallen - I haven't been able to run the GN unit tests with HNAP installed for a very long time (I dont think ever).  This is an architectual issue in GN with how it manages schemas and dependencies - its a bit of a mess as you're seen.

However, you should be able to resolve it by adding HNAP as a dependency (test scope) to the project that has failing test cases.

ianwallen: @davidblasby good idea about adding the dependency.  I have created a new PR that will do that. see https://github.com/geonetwork/core-geonetwork/pull/4835

ianwallen: @davidblasby looks like there was another bug with the spring files once I added the dependency.  I created pull request https://github.com/metadata101/iso19139.ca.HNAP/pull/61 to correct this issue.

PR #61, opened 2020-07-08
Fix bug in spring file.

ianwallen:
It was generating the following error.

org.springframework.beans.factory.xml.XmlBeanDefinitionStoreException: Line 10 in XML document from URL [file:/C:/Projects/TADAP/geonetwork-dev/schemas/iso19139.ca.HNAP/target/classes/config-spring-geonetwork.xml] is invalid; nested exception is org.xml.sax.SAXParseException; lineNumber: 10; columnNumber: 75; cvc-complex-type.2.4.c: The matching wildcard is strict, but no declaration can be found for element 'context:component-scan'.

This was caused when trying to run unit test from service module GetKeywordByIdTest

See discussion https://github.com/metadata101/iso19139.ca.HNAP/pull/56#discussion_r450872749 more details

### DIFF
--- a/src/main/resources/config-spring-geonetwork.xml
+++ b/src/main/resources/config-spring-geonetwork.xml
@@ -3,9 +3,11 @@
   xmlns="http://www.springframework.org/schema/beans"
   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
   xmlns:util="http://www.springframework.org/schema/util"
-  xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd"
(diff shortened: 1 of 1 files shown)


ianwallen:
@davidblasby, can you review these changes.  It seems that you added these changes with commit https://github.com/metadata101/iso19139.ca.HNAP/commit/4569226671ca386f47c8a81138eaec3bbc7eef78

Note: I do not fully understand the error - I found a similar configuration using context:component-scan and copied it to this file and it fixed the problem.

Thank you

davidblasby:
@ianwallen  - I will look at it in the morning.  I don't think there should be a problem.  I'll just make sure its finding the beans.

davidblasby (approved):
Looks good

PR #62, opened 2020-07-09
Add import converter for converting iso19139 to HNAP

ianwallen:
I attempted to export and import some data from FGP and got some errors in the import.

I received some data which contains the following.
```
<gmd:metadataStandardName xsi:type="gmd:PT_FreeText_PropertyType">
    <gco:CharacterString>North American Profile of ISO 19115:2003 - Geographic information - Metadata</gco:CharacterString>
    <gmd:PT_FreeText>
      <gmd:textGroup>
        <gmd:LocalisedCharacterString locale="#fra">fra</gmd:LocalisedCharacterString>
      </gmd:textGroup>
    </gmd:PT_FreeText>
  </gmd:metadataStandardName>
  <gmd:metadataStandardVersion>
    <gco:CharacterString>NAP - CAN/CGSB-171.100-2009</gco:CharacterString>
  </gmd:metadataStandardVersion>

```

`NAP - CAN/CGSB-171.100-2009`  should be `CAN/CGSB-171.100-2009`
And this was causing the schema detection to detect this as a ISO19139.

I also received other records where the autodetect was failing to pick HNAP due to the metadataStandard elements not being set correctly.  As we have our system setup to restrict the schema's to HNAP only we could not load the metadata records without fixing the problem.

1 - it was difficult to identify the records from the MEF file that were causing issued.
Created pull request to give better error message.
https://github.com/geonetwork/core-geonetwork/pull/4838

2 - It was time consuming to identify and fix each error.
So I created a converter that would modify the metadataStandard tags to make it HNAP compliant so that the autodetect would not choose the wrong schema.  That is what this pull request is for.

The converter will automatically install itself as part of the schema initialization process.

### DIFF
--- a/src/main/java/ca/gc/schemas/iso19139hnap/init/SchemaInitializerConverters.java
+++ b/src/main/java/ca/gc/schemas/iso19139hnap/init/SchemaInitializerConverters.java
@@ -44,7 +44,7 @@ public class SchemaInitializerConverters implements
             Path correspondingDataDirFile = Paths.get(conversionPath.toString(), dir, fname);
             if (!Files.exists(correspondingDataDirFile)) {
                 //need to copy it in...
-                Log.info(Geonet.SCHEMA_MANAGER, "ISO19139.HNAP: SchemaInitializer: copying conversion files: " + fname + " to " + correspondingDataDirFile.toString());
+                Log.info(Geonet.SCHEMA_MANAGER, "ISO19139.HNAP: SchemaInitializerConverters: copying conversion files: " + fname + " to " + correspondingDataDirFile.toString());
                 try (InputStream is = r.getInputStream()) { //auto close
                     Files.copy(is, correspondingDataDirFile);
                 }
@@ -60,7 +60,7 @@ public class SchemaInitializerConverters implements
         try {
             ConfigureThesauruses(conversionPath);
         } catch (IOException e) {
-            Log.error(Geonet.SCHEMA_MANAGER, "SchemaInitializer: preloading conversion files", e);
+            Log.error(Geonet.SCHEMA_MANAGER, "SchemaInitializerConverters: preloading conversion files", e);
         }
     }
 }
--- a/src/main/java/ca/gc/schemas/iso19139hnap/init/SchemaInitializerConverters.java
+++ b/src/main/java/ca/gc/schemas/iso19139hnap/init/SchemaInitializerConverters.java
@@ -1,0 +1,66 @@
+package ca.gc.schemas.iso19139hnap.init;
+
+import org.fao.geonet.constants.Geonet;
+import org.fao.geonet.kernel.GeonetworkDataDirectory;
+import org.fao.geonet.utils.Log;
+import org.springframework.context.ApplicationListener;
+import org.springframework.core.io.Resource;
+import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
+
+import javax.annotation.PostConstruct;
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.regex.Pattern;
+
+/**
+ * This class is for setting up a schema conversion files
+ * <p>
+ * It watches the system startup and gives us the opportunity to do any setup required at the appropriate time.
+ * <p>
+ * See config-spring-geonetwork.xml in this schema
+ */
+public class SchemaInitializerConverters implements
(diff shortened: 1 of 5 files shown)


ianwallen:
@davidblasby, the SchemaInitializerConverters.java  was mostly a copy from your thesaurus code.

I'm not sure if putting the files in xsl/conversion (which matches the webapp folder) is a good place to put the converters?

src/main/java/ca/gc/schemas/iso19139hnap/init/SchemaInitializerConverters.java
```diff
@@ -0,0 +1,66 @@
...
+            String[] subDirs = r.getURI().toURL().getPath().split(Pattern.quote("/"));
+            String dir = subDirs[subDirs.length - 2]; // ie. "import"
+            ///put in - xsl/conversion/import/iso19139-to-iso19139.ca.HNAP.xsl
+            //ensure that dirs exist
+            Files.createDirectories(Paths.get(conversionPath.toString(), dir));
+            //full path to put the .xsl
+            Path correspondingDataDirFile = Paths.get(conversionPath.toString(), dir, fname);
+            if (!Files.exists(correspondingDataDirFile)) {
+                //need to copy it in...
+                Log.info(Geonet.SCHEMA_MANAGER, "ISO19139.HNAP: SchemaInitializer: copying conversion files: " + fname + " to " + correspondingDataDirFile.toString());
```
davidblasby: might want to change this to "SchemaInitializerConverters" just for clarity...

src/main/java/ca/gc/schemas/iso19139hnap/init/SchemaInitializerConverters.java
```diff
@@ -0,0 +1,66 @@
...
+
+
+    //when the data dir is configured, we're good to copy in our conversion files
+    @Override
+    public void onApplicationEvent(GeonetworkDataDirectory.GeonetworkDataDirectoryInitializedEvent event) {
+        Path conversionPath = event.getSource().getWebappDir().resolve(conversionDir);
+        try {
+            ConfigureThesauruses(conversionPath);
+        } catch (IOException e) {
+            Log.error(Geonet.SCHEMA_MANAGER, "SchemaInitializer: preloading conversion files", e);
```
davidblasby: might want to change this to "SchemaInitializerConverters" just for clarity...

davidblasby (commented):
@ianwallen  - I think the xslt would better to be put in plugin/iso19319.ca.HNAP/convert/

I think keeping all the XSLT inside the schema will make things easier, long term.

ianwallen:
> @ianwallen - I think the xslt would better to be put in plugin/iso19319.ca.HNAP/convert/
> 
> I think keeping all the XSLT inside the schema will make things easier, long term.

@davidblasby, I don't quite understand your comment.  The xslt are inside the schema.   The src/main/config/conversion/* will be copies to the webapp/xsl/conversions/* and in this case src/main/config/conversion/import/iso19139-to-iso19139.ca.HNAP.xsl just contains an xsl:include to iso19139.ca.HNAP/convert/ISO19139/fromISO19139.xsl which is in the convert folder that you mentioned.  As  src/main/config/conversion/* is installed outside the schema plugin (similar to the thesauri init that you did), I placed the files in the same src/main/config folded and my quesion was mostly asking if this was a good place to put files that will be copied outside the plugin folder.

Could you provide an example of what specific file should be moved inside plugin/iso19319.ca.HNAP/convert/ folder?

Note: I took the instructions for copying the conversion files to the target location from the following link.
https://github.com/metadata101/iso19115-3.2018#adding-the-conversion-to-the-import-record-page
(I think it would be good if iso19115-3.2018 used this logic for copying the files automatically instead of a manual process)

davidblasby:
@ianwallen  - ok, maybe i was misunderstanding.  Looks good!

josegar74:
@davidblasby the xslt can't be really added to the schema as in the import process potentially you can import metadata from different schemas. For this, are stored globally in `webapp/xsl/conversions`.

fxprunayre:
Maybe it would make sense to move all conversions to their schemas and then group them in the import panel ?

image

and a future improvement would be to have a label instead of filename.

ianwallen:
@fxprunayre  great suggestions - I have created a new enhancement issue - https://github.com/geonetwork/core-geonetwork/issues/4852

ianwallen:
As https://github.com/geonetwork/core-geonetwork/issues/4852 is currently not implemented and there has not been any objections to this pull request, I will merge the pull request for now.  If/when https://github.com/geonetwork/core-geonetwork/issues/4852 is done, I will revert some of the changes from this pull request.

PR #65, opened 2020-07-10
Store index format and orgName

ianwallen:
Store `format` in index so that it is returned in the query interface.

Update: added `orgName` as well


ianwallen:
@josegar74 do you know why these were set to store=false?  Do you have any objections to changing these to store=true like all the other plugins are doing?

ianwallen:
There has been no objections - I plan on merging these changes into the branch tomorrow unless someone objects.

PR #73, opened 2020-08-11
Replaced "bold" class name with WET standard "h6" class name

bo-lu:


src/main/plugin/iso19139.nap/formatter/ec-view/view.xsl
```diff
@@ -393,7 +393,7 @@
           <xsl:variable name="geom" select="concat('POLYGON((', $minx, ' ', $miny,',',$maxx,' ',$miny,',',$maxx,' ',$maxy,',',$minx,' ',$maxy,',',$minx,' ',$miny, '))')"/>
 
           <xsl:if test="$hasGeometry">
-            <span class="bold"><xsl:value-of select="$schemaStrings/GeographicExtent"/></span><br/>
+            <span class="h6"><xsl:value-of select="$schemaStrings/GeographicExtent"/></span><br/>
```
jodygarnett, line 396: This looks good, thanks @bo-lu

jodygarnett (approved):

src/main/plugin/iso19139.nap/formatter/ec-view/view.xsl
ianwallen: @jodygarnett 
Should this change be applied to 3.10.x as well?

Issue #74, opened 2020-08-12
Add bilingual EN/FR Titles to RDF's


geothorne:
Once Rdf's are renamed #53 @davidblasby if you are available can you add the following titles:
 
rdf-identifier | en-Title | fr-Title
-- | -- | --
EC_Branch | Branch | Direction Générale
GC_Core_Subject | GC Core Subject | Sujet central de la GC
EC_Directorate | Directorate | Direction
GC_Security_Level | GC Security Level | GC Restrictions de manipulation
GC_Geographic_Scope | Geography | Portée Géographique
GC_Departments | Government of Canada Departments & Agencies | Ministères et organismes du gouvernement du Canada
ISO_Countries | Countries | Pays
EC_Program_PAA | Program (PAA) | Portée du programme (PAA)
EC_Functions | Business Functions | Fonctions de l'entreprise
NE_Global_Regions | Global Regions | Régions mondiales
GC_State_Province | Canada/US States and Provinces | Canada/US États et provinces
GC_Resource_Format | Resource Format | Format de fichier

davidblasby:
These as mentioned in #53 these are not in the metadata101/hnap repo;

EC_Branch ---> not in metadata101/HNAP
EC_Directorate ---> not in metadata101/HNAP
GC_Security_Level ---> not in metadata101/HNAP
EC_Program_PAA ---> not in metadata101/HNAP
NE_Global_Regions ---> not in metadata101/HNAP

davidblasby:
https://github.com/metadata101/iso19139.ca.HNAP/pull/75

geothorne:
[missing-rdfs.zip](https://github.com/metadata101/iso19139.ca.HNAP/files/5070501/missing-rdfs.zip)

@davidblasby 
Here are all but NE_Global_Regions (regions.rdf).
As you mentioned in #53 

This is part of CORE GeoNetwork - https://github.com/geonetwork/core-geonetwork/blob/master/web/src/main/webapp/WEB-INF/data/config/codelist/external/thesauri/place/regions.rdf

I currently don't see value in making a duplicate copy. Thanks!

davidblasby:
Added thesauri - https://github.com/metadata101/iso19139.ca.HNAP/pull/78

jodygarnett:
I think this issue may be resolved

PR #54, opened 2020-06-12
Removing unnecessary EC thesaurus

geothorne:
As per issue 53: https://github.com/metadata101/iso19139.ca.HNAP/issues/53

EC has done a review of ECCC current thesaurus and have identified several rdf that not longer valid. 

* [x] EC_Content_Scope
* [x] EC_Data_Usage_Scope
* [x] EC_Government_Names
* [x] EC_Information_Category
* [x] EC_Org_Names
* [x] EC_Resource_ContentTypes

Issue #53, opened 2020-06-06
Rename non EC* thesaurus files to GC* 


ianwallen:
I believe we should rename the none ECCC specific thesaurus files to GC. 

i.e. I don't believe the `EC_States.rdf` is specific to ECCC and can probably be applied to all government departments so it may be better to have it renamed to `GC_States.rdf`.  

As I don't know all the uses for the thesaurus, I believe this should be an ECCC task.

jodygarnett:
Is it useful to include all the remaining EC* thesaurus in the schema plugin, or should we isolate into a profile (as per https://github.com/metadata101/iso19139.ca.HNAP/issues/45 )

ianwallen:
I don't have any issues/objections  if EC* remains in HNAP for now and maybe moved later as per #45. Just trying to clarify the current ones so that we know which one are GC and which ones are EC.

jodygarnett:
@geothorne I believe you are looking at simplifying `rdf` files, can we look at this request at the same time.

geothorne:
@davidblasby please help **Rename** the following RDF's:

**Organizational Prefix's:**
_GC -> Government of Canada_
_ISO -> iso.org_
_NE -> Natural Earth - www.naturalearthdata.com_
_EC -> Environment and Climate Change Canada (Environment Canada)_

**Suffix Acronyms:**
_PAA -> Program Alignment Architecture_
- [x] 1. EC_Core_Subject -> GC_Core_Subject
- [ ] 2. EC_GC_Security_Classification -> GC_Security_Level
- [x] 3. EC_Geographic_Scope -> GC_Geographic_Scope
- [x] 4. EC_Government_Titles -> GC_Departments 
       _(I believe this file is managed by Federal Geospatial Platform (FGP) for their Catalogue)_
- [x] 5. EC_ISO_Countries -> ISO_Countries
- [ ] 6. EC_Program -> EC_Program_PAA
- [x] 7. EC_WAF -> EC_Functions
- [ ] 8. regions -> NE_Global_Regions
- [x] 9. EC_States -> GC_State_Province 
- [x] 10. EC_Resource_Formats  -> GC_Resource_Format

Note: The following rdf's do not change name
11. EC_Directorate
12. EC_Branch

davidblasby:
The following are unchanged;

EC_Content_Scope.rdf
EC_Data_Usage_Scope.rdf
EC_Information_Category.rdf
EC_Org_Names
[...]

PR #75, opened 2020-08-12
rename thesaurus, remove unused thesaurus, and add eng/fre names

davidblasby:
Renamed and added English/French titles to some of the thesauri.

See #74 #53 #54


davidblasby:
@geothorne  - please ensure correct
@ianwallen  - FYI on these changes.  See the three tickets #74  #53  #54

ianwallen:
Is this PR ready?

I did a search for "external.theme"   and found the following 

- 'external.theme.EC_Resource_Formats' - should be rename?
- "thesaurus": "external.theme.EC_Org_Names",  - no longer exists.?
- "external.theme.EC_ISO_Countries" - should be renamed?
- etc...

If we rename EC_ISO_Countries to ISO_Countries then it is not so much hnap specific and I can see the possibility of geonetwork core using this name in the future and causing a conflict.  Do you think it would make sense to contribute this to Geonetwork core?   This is just a concern - I will still accept the PR if this is not changed.

davidblasby:
@geothorne @ianwallen  - added in 2 deleted thesauri that were used by either the Organization Selector or Validation (see ticket #54 )
Changed references to the thesauri

ianwallen (approved):

PR #80, opened 2020-08-25
Index gmd:geometricObjectType values

josegar74:


src/main/plugin/iso19139.ca.HNAP/index-fields/default.xsl
```diff
@@ -620,6 +620,23 @@
       </xsl:for-each>
     </xsl:for-each>
 
+    <xsl:for-each select="gmd:spatialRepresentationInfo/gmd:MD_VectorSpatialRepresentation/gmd:geometricObjects/gmd:MD_GeometricObjects/gmd:geometricObjectType/gmd:MD_GeometricObjectTypeCode">
+      <xsl:variable name="geometryTypeValue">
+        <xsl:choose>
+          <xsl:when test="@codeListValue = 'RI_508'">point</xsl:when>
+          <xsl:when test="@codeListValue = 'RI_507'">line</xsl:when>
+          <xsl:when test="@codeListValue = 'RI_509' or @codeListValue = 'RI_510'">area</xsl:when>
+          <xsl:when test="@codeListValue = 'RI_505' or @codeListValue = 'RI_506'">multiple></xsl:when>
```
ianwallen, line 635: Do we want to hard code values from the codelist instead of referencing the values from
https://github.com/metadata101/iso19139.ca.HNAP/blob/3.10.x/src/main/plugin/iso19139.ca.HNAP/loc/eng/codelists.xml#L1106-L1154

Seems odd to attempt to group some of the codes in "multiple" and "area" instead of using te original values.  If the original values are incorrect then maybe the codelist needs to change?

Also it does not seem to support multilingual which I would expect to be required?

davidblasby: I see all 6 values (505 to 510) are used - good.

Not sure what the end result of this is trying to do - I see it creates a Field (line 636) of type geometryValue.  Does this end up in the resulting Record?

josegar74: This is used in a project to display an icon indicating the type of geometry, so the value it's used to select a CSS value, that's why not used the original values in English and French and also grouped to simplify the symbology.

ianwallen: @josegar74 
OK - that makes more sense.
I think if you apply a comments indicating that the code is doing a remapping to CSS values then it should be ok.

davidblasby: ok, I understand!

davidblasby (approved):

PR #82, opened 2020-08-25
minor updates to make org selector more configurable

davidblasby:
1. I added Ordered to the ThesaurusInitizer for listening to GeonetworkDataDir events.
    * this is to make it easier for downstream projects (i.e. municipalities/other orgs) to do things with the thesaurus

2. I made it so the org selector's top-level item (current Government of Canada) can be set with Settings 
     * added default setting
     * added text for the setting   
     * use the setting in the schema

### DIFF
--- a/src/main/config/translations/en-schema-iso19139.ca.HNAP.json
+++ b/src/main/config/translations/en-schema-iso19139.ca.HNAP.json
@@ -1,5 +1,8 @@
 {
   "schema": "Schema-Specific Settings",
   "schema/iso19139.ca.HNAP": "ISO19139 HNAP Settings",
-  "schema/iso19139.ca.HNAP/UseGovernmentOfCanadaOrganisationName": "Use 'Government of Canada' Organization Selector"
+  "schema/iso19139.ca.HNAP/UseGovernmentOfCanadaOrganisationName": "Use 'Government of Canada' Organization Selector",
+  "schema/iso19139.ca.HNAP/DefaultMainOrganizationName_en": "Default top-level Organization Name (English)",
+  "schema/iso19139.ca.HNAP/DefaultMainOrganizationName_fr": "Default top-level Organization Name (French)"
+
 }
--- a/src/main/config/translations/en-schema-iso19139.ca.HNAP.json
+++ b/src/main/config/translations/en-schema-iso19139.ca.HNAP.json
@@ -1,8 +1,8 @@
 {
   "schema": "Schema-Specific Settings",
   "schema/iso19139.ca.HNAP": "ISO19139 HNAP Settings",
-  "schema/iso19139.ca.HNAP/UseGovernmentOfCanadaOrganisationName": "Use 'Government of Canada' Organization Selector",
-  "schema/iso19139.ca.HNAP/DefaultMainOrganizationName_en": "Default top-level Organization Name (English)",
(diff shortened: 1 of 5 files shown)


ianwallen:
@davidblasby 
Based on the changes, it seems odd that the text has changed from "Government of Canada" to "Default top-level Organization Name (English)" for en.

I also find it odd to have french reference "DefaultMainOrganizationName_fr" in and english file "en-schema-iso19139.ca.HNAP.json" and vice versa for the french.  

Could you tell me where these changes will be visible in the interface.  Maybe once I see where these changes are visible, it may make more sense?

davidblasby:
English UI (admin/settings):

image

French UI (admin/settings):

image


@ianwallen  - its at the very top of the settings (administration).

Those two values are the default for the top-level (currently "Government of Canada"/"Gouvernement du Canada") for organization name in the editor;
image

This is so that you could have different organizations (i.e. not the government of canada) use the controller.  Its setup so that the default HNAP hasn't changed.  However, different distributions can (a) use different GC_Org_Name.rdf to control this list and (b) have a different default organization name.

Does that make sense?

ianwallen:
@davidblasby 
Thank you for the clarification.

Odd that your french screenshot contains English text for the french version.  The text was correct on my system. - I'm assuming this was an older screenshot?

It would be better if the language values would be similar to the editor and use the English French switch/tab instead of 2 strings.  By using the language selected it would also be cleaner when adding more languages.  If this interface used more than 2 languages, it would be increasingly difficult to manage.
image

As you changed the context, I'm not sure that the following text is still correct.
`Use 'Government of Canada' Organization Selector`
It seems odd to say "Use 'Government of Canada' Organization Selector" and then put "City of Victoria". I'm assuming that if this is used for the City of Victoria then the organization selector thesaurus file will also be changed to reflect the smaller divisions of the city?  If this is the case then it seems that what this option is really doing is selecting the thesaurus to use for the org selector.  And maybe the thesaurus can include the default in there? Maybe that is a more complicated change at this time?

davidblasby:
Yes - I noticed the mistake and fixed it (but didn't update the screenshot).  Opps.

I agree - if there were a bunch of languages, the settings-config would be more complex.  However, its not obvious how to correct this without major changes.  Since HNAP is just eng/fra I don't think this needs to be changed at the moment?

I will change the "Use 'Government of Canada' Organization Selector" and some of the other text to make it more obvious what it's doing.

Yes - the new deployment system (we talked about today) replaces the GC_Org_Names and GC_Departements thesaurus with city-specific thesauri.  

We could try to add something to the RDF to indicate the default - not quite sure if that would be spec-compliant but shouldn't be too difficult.  However, let's leave that to the future as its going to cause a cascade of changes.

ianwallen (approved):
For the language, I believe a change would need to be done to geonetwork core to support multilingual config values.  So I think it would be easier to look into the default value in RDF first. But as you said, this can be a future task.

josegar74 (commented):
@davidblasby, it works fine for me, but for non-multilingual template I see the english language pill for the branch field:

non-multilingual

It doesn't affect the behaviour, but if can be hidden could be better.

davidblasby:
@josegar74  - thats already fixed in https://github.com/geonetwork/core-geonetwork/pull/4964

josegar74 (approved):

PR #85, opened 2020-08-27
Add EPSG:26917 and EPSG:3857 to labels.xml

jodygarnett:


davidblasby (approved):

jodygarnett:
@ianwallen is it okay to add two more codes here? I also could not figure out why 4326 was labeled incorrectly?

ianwallen:
@jodygarnett 

Regarding EPSG:26917

According to nrcan
https://www.nrcan.gc.ca/earth-sciences/geography/topographic-information/maps/utm-grid-map-projections/utm-grid-universal-transverse-mercator-projection/9779

Canada uses UTM zone 7 to 22.  So I'm a little confused as to why you are only adding zone 17 and not all the others?

jodygarnett:
> Regarding EPSG:26917
> Canada uses UTM zone 7 to 22. So I'm a little confused as to why you are only adding zone 17 and not all the others?

I think other UTM zones have other numbers? I am basing this off a quick lookup of

* https://epsg.io/26917
* https://spatialreference.org/ref/epsg/26917/
* https://epsg.org/crs_26917/NAD83-UTM-zone-17N.html

The last one is the official EPSG database, which apparently has a new home.

I think NRCan uses several UTM zones, and thus a range of EPSG codes. I guess I was just listing this one as it is the one I needed for a customer. Should we add the complete set now, or just as needed?

jodygarnett:
So those zones would cover:

* 7N: https://epsg.io/26907
* ...
* 15N: https://epsg.io/26915
* 16N: https://epsg.io/26916
* ...
* 21N:  https://epsg.io/26921

ianwallen:
I have no objection adding the extra ones to the list.  As this is mostly a Canadian plugin, we should probably add them.
We just need to be careful that this list does not get too long/complicated....

PR #86, opened 2020-09-07
Validation rules refactor

josegar74:
1. Split validation rules in 3 files:

- `schematron-rules-common.sch`: codelist validations and text fields that are not intended to be multilingual (for example, distribution format).
- `schematron-rules-multilingual`: validation rules for multilingual fields in multilingual metadata.
- `schematron-rules-non-multilingual`: validation rules for multilingual fields in non-multilingual metadata.

2. Created a vocabulary for licenses and use it in the validation rules instead of hardcoding license values.

3. Improve validation message codes.

4. Cleanup unused validation files.

---

To configure the schematron `schematron-rules-multilingual` to be executed only for multilingual metadata:
multilingual-rules

To configure the schematron `schematron-rules-non-multilingual` to be executed only for non-multilingual metadata:

non-multilingual-rules


jodygarnett:
request: For the requirement to use an open license, I would like to do this check in a separate schematron file (so it can be disabled when working with regional data).

ianwallen:
Tested and it seems to work fine. No objections on the changes.
Only suggestion would be to have multilingual and unilingual instead of non-multilingual.  Just a suggestion - not a requirement.

In our case, we will be publishing to FGP and Open Data so we would like to have an HNAP Level 1 validation (maybe level 2 and 3 as well - if possible) and open data as well.  I suspect that these rules would include both schematron-rules-multilingual and schematron-rules-common.  (open data would not require the geospatial requirement).   For end users I believe I believe it may be too confusing to have them know that FGP requires common and multilingual... so in the end I'm not sure if we would be using the common/multilingual and instead we may simply create new one. We will probably be looking into this in the near future.

ianwallen (approved):

PR #91, opened 2020-09-22
3.6.x translation fixes

bo-lu:
Simplified the word 'Collection' and fixed buttons that weren't translated in French


jodygarnett (approved):
This looks good to me

Issue #99, opened 2020-10-20
Org selected - incorrect error/too many errors


ianwallen:
The following 

![image](https://user-images.githubusercontent.com/1868233/96596012-a3715100-12c2-11eb-8f24-999532fb463b.png)

Generates the following error.

![image](https://user-images.githubusercontent.com/1868233/96595943-90f71780-12c2-11eb-9e75-be68d8648646.png)

There are a couple issues.
1 - The org name is correct and the error is incorrect.
2 - If the org name is not correct it should only show one error not 2.  Once the first issues is corrected then we can show the second.  They are sort of dependent.

PR #100, opened 2020-10-20
Fix issue with org name validation messages

ianwallen:
Fix for #99

Fixed bug with checking organization name - gov of canada as it was looking the name with a ";" at the end. But if the department/title was not filled in there would not be a ";". So changed to do the check by checking the first token of a ";" separated string.

Fixed bug with compound messages. i.e. if the gov of canada had errors then all the other validation would also fail. So it could generate 4 error and this was making the errors confusing.  So changes it so that it only generated up to 1 error for each org check.


src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
```diff
@@ -106,7 +106,7 @@
       <sch:let name="government-names" value="document(concat('file:///', replace(concat($thesaurusDir, '/external/thesauri/theme/GC_Org_Names.rdf'), '\\', '/')))"/>
 
       <sch:let name="organisationName" value="gco:CharacterString" />
-      <sch:let name="isGovernmentOfCanada" value="starts-with(lower-case($organisationName), 'government of canada;') or starts-with(lower-case($organisationName), 'gouvernement du canada;')" />
+      <sch:let name="isGovernmentOfCanada" value="starts-with(lower-case(normalize-space(tokenize($organisationName, ';')[1])), 'government of canada') or starts-with(lower-case(normalize-space(tokenize($organisationName, ';')[1])), 'gouvernement du canada')" />
```
jodygarnett, line 109: This should be the value from the settings `schema/iso19139.ca.HNAP/DefaultMainOrganizationName_fr` nad `schema/iso19139.ca.HNAP/DefaultMainOrganizationName_en`, rather than hard coded to "government of canada".

ianwallen: @jodygarnett 
I don't disagree with this request however it is not part of this PR.  

I'm also not sure how to import these variables from layout/custom-fields.xsl  or from the settings as /root/gui/settings/schema/iso19139.ca.HNAP/DefaultMainOrganizationName_en

I suggest that you create a new issue for this request.

Issue #97, opened 2020-10-20
Interesting challenge for non GC_Open_license.rdf use


jodygarnett:
I have an interesting challenge for non GC use, the file [schematron-rules-non-multilingual.sch#L136-L151](https://github.com/metadata101/iso19139.ca.HNAP/blob/3.10.x/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-non-multilingual.sch#L136-L151)

```xml
    <!-- Use Limitation -->
    <sch:rule context="//gmd:identificationInfo/gmd:MD_DataIdentification
            |//*[@gco:isoType='gmd:MD_DataIdentification']
            |//*[@gco:isoType='srv:SV_ServiceIdentification']">

      <sch:let name="open-licenses" value="document(concat('file:///', replace(concat($thesaurusDir, '/external/thesauri/theme/GC_Open_licenses.rdf'), '\\', '/')))"/>

      <sch:let name="openLicense" value="count(gmd:resourceConstraints/gmd:MD_LegalConstraints/gmd:useLimitation[
               (normalize-space(gco:CharacterString) = $open-licenses//rdf:Description/ns2:prefLabel[@xml:lang=$mainLanguage2char])])" />


      <sch:assert
        test="$openLicense > 0"
      >$loc/strings/OpenLicense</sch:assert>

    </sch:rule>
```

The above rule wants to ensure the record is available as one of the approved open data licenses defined by `GC_Open_licenses.rdf` thesaurus.

Options:

a) Split this into a seperate schematron file, so the check can be disabled as a configuration option

b) Replace `GC_Open_licenses.rdf` with an appropriate license

c) change the `$openLicense` check

d) can we make a configuration options (to send into the schematron)

ianwallen:
I would expect that you would have similar issues with the organization names and other thesaurus?
As this organization is not GC - it sounds like they should be using their own thesaurus - So maybe the schematron needs to be modified so support an option to supply the thesaurus to use.  
It would be similar to option "b)" but instead of replacing GC_Open_licenses.rdf and putting data that is not GC open licenses, it would instead point to Other_Open_licenses.rdf which makes more sense and may avoid future confusion.

jodygarnett:
You can see an experiment here - https://github.com/GeoCat/experiment-hnap
Right now the war overlay over-writes those two files.

We should consider adding options as you describe longer term.

jodygarnett:
Option (b) is working, going to mark this as addressed.

PR #101, opened 2020-10-20
Factor out open license check

jodygarnett:
In response to https://github.com/metadata101/iso19139.ca.HNAP/issues/97

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-open-license.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-open-license.sch
@@ -1,0 +1,98 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<sch:schema xmlns:sch="http://purl.oclc.org/dsdl/schematron"
+            xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+            xmlns:geonet="http://www.fao.org/geonetwork"
+            xmlns:xs="http://www.w3.org/2001/XMLSchema"
+            queryBinding="xslt2">
+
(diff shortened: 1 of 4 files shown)


src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-open-license.sch
```diff
@@ -0,0 +1,98 @@
...
+    <xsl:sequence
+            select="
+            every $listValue in $list
+            satisfies exists($searchStrings[. = $listValue])
+            "
+    />
+  </xsl:function>
+
+  <!-- ===================================================================
+  EC schematron rules for open-data license validation in metadata editor:
```
ianwallen, line 68: Don't believe this is and EC rule.

ianwallen (commented):
How is this fixing your issue where another agency is going to have its own license?  Are you planning on creating another license schematron for them?

I know from our end, management does not like having all these schematrons as they feel it confuses the users.   They would rather have validations required for the purpose. 
i.e. we will be publishing data to FGP, Open Data, internal Internal... And the users may only know English. The reason for language specific is because they will send the other language for translation only after the initial language passes validation.
So they would like the users to validate FGP English, FGP French, Open Data English.... 
But this can end up with a lot of schematrons.... 

Lets bring this topic up at the next meetings

jodygarnett:
Is there an include? That way we could make one check for each service .. which includes only the parts needed.

jodygarnett:
Closing this, overriding the `rdf` file contents is working out for now.

I would still like to pursue the idea of using xml import to make these checks match different publication targets.

PR #104, opened 2020-10-23
Fix french index on keywords

ianwallen:
The english facet has the following on our system

image

While the french page has the following.

image

The french is missing keywords and formats among others.

The french index was using FRA instead of FRE so it was not located the data.

This PR also applies some of the latest changes from the iso19139 mostly regarding the keywords. (This was done in the first commit while attempting to locate the real issue).  The bug is really fixed in the second commit.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/index-fields/language-default.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/index-fields/language-default.xsl
@@ -84,7 +84,14 @@
         <xsl:variable name="isoLangId"
                       select="normalize-space(string(gmd:languageCode/gmd:LanguageCode/@codeListValue))"/>
         <xsl:if test="$isoLangId!=$isoDocLangId">
-          <Document locale="{$isoLangId}">
+          <!-- get iso language code as ISO639 2B -->
+          <xsl:variable name="isoLangId_ISO639_2B">
+            <xsl:choose>
+              <xsl:when test="$isoLangId = 'fra'">fre</xsl:when>
+              <xsl:otherwise>$isoLangId</xsl:otherwise>
+            </xsl:choose>
+          </xsl:variable>
+          <Document locale="{$isoLangId_ISO639_2B}">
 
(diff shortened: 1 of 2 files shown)


src/main/plugin/iso19139.ca.HNAP/index-fields/language-default.xsl
```diff
@@ -84,7 +84,14 @@
...
         <xsl:if test="$isoLangId!=$isoDocLangId">
-          <Document locale="{$isoLangId}">
+          <!-- get iso language code as ISO639 2B -->
+          <xsl:variable name="isoLangId_ISO639_2B">
+            <xsl:choose>
+              <xsl:when test="$isoLangId = 'fra'">fre</xsl:when>
+              <xsl:otherwise>$isoLangId</xsl:otherwise>
+            </xsl:choose>
+          </xsl:variable>
+          <Document locale="{$isoLangId_ISO639_2B}">
```
jodygarnett, line 94: So this is the specific change being worked on changing `fra` to `fre`

jodygarnett (approved):
I can follow the specific `fra` / `fre` change

josegar74 (approved):
Tested, working fine. Code looks ok.

Something to check is related to the code related to INSPIRE as only applies in Europe, can be removed safely. But can be done in another pull request.

ianwallen:
@josegar74  The problem is that ISO19139 is getting updates and we eventually need to merge some of these updates back into this schema.  I find it fairly difficult to identify what fixes need to be applied and what changes were done specifically for HNAP.

So for that reason, I don't mind having the INSPIRE code there when it makes it easier to compare the changes between ISO19139 and this one.  As the INSPIRE flags are false, it should not use any of the code anyway.  I will generally also adjust the formatting so that it matches the ISO10139 files which also makes it easier to compare.

It would be better if we could just include the ISO19139 and just make HNAP specific changes. But this is not always easy.

If you are aware of better ways to synchronize the changes between ISO19139 and HNAP, I would be interested in hearing this process.

PR #105, opened 2020-10-23
Fixed bug where wrong protocol dropdown list used

ianwallen:
Fixed bug where wrong protocol dropdown list was used. It was using the one from iso19139

Also added a rest service to the list.


jodygarnett (approved):

ianwallen:
partially fixes #106

PR #117, opened 2020-11-04
Fixed bug with language retrieval/usage in schematron

ianwallen:
Fixed bug with language retrieval.  The main language will not always have a local with a languageid so languageid can be null and should not be used then value is required.

For same reason fix metadatastandard checks that were using mainlanguageid instead of mainlanguage - otherwise the check could fail.

Also fixed EC19 which was renamed to MetadataStandardName - message was not being displayed.

Without this fix we were getting the following error when attempting to import and validate some records that did not have the local for main language
image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-non-multilingual.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-non-multilingual.sch
@@ -22,14 +22,14 @@
   <sch:ns prefix="rdfs" uri="http://www.w3.org/2000/01/rdf-schema#"/>
 
   <sch:let name="schema" value="'iso19139.ca.HNAP'"/>
-  <sch:let name="mainLanguage" value="if (normalize-space(//*[name(.)='gmd:MD_Metadata' or @gco:isoType='gmd:MD_Metadata']/gmd:language/gmd:LanguageCode/@codeListValue) != '')
-                                       then normalize-space(//*[name(.)='gmd:MD_Metadata' or @gco:isoType='gmd:MD_Metadata']/gmd:language/gmd:LanguageCode/@codeListValue)
+<sch:let name="mainLanguage" value="if (normalize-space(//*[name(.)='gmd:MD_Metadata' or @gco:isoType='gmd:MD_Metadata']/gmd:language/gmd:LanguageCode/@codeListValue) != '')
+                                       then lower-case(normalize-space(//*[name(.)='gmd:MD_Metadata' or @gco:isoType='gmd:MD_Metadata']/gmd:language/gmd:LanguageCode/@codeListValue))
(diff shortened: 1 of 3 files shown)


src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-non-multilingual.sch
```diff
@@ -72,8 +72,8 @@
     <!-- Metadata Standard Name -->
     <sch:rule context="//gmd:metadataStandardName">
 
-      <sch:let name="correct" value="(($mainLanguageId = 'eng' and gco:CharacterString = 'North American Profile of ISO 19115:2003 - Geographic information - Metadata') or
-                          ($mainLanguageId = 'fra' and gco:CharacterString = 'Profil nord-américain de la norme ISO 19115:2003 - Information géographique - Métadonnées'))
+      <sch:let name="correct" value="(($mainLanguage = 'eng' and gco:CharacterString = 'North American Profile of ISO 19115:2003 - Geographic information - Metadata') or
```
jodygarnett, line 75: Above `lower-case($mainLanguage)` is used for this test, do we need to be consistent for this to work?

josegar74: I would apply `lower-case` to the `select` expression in `$mainLanguage` to avoid having to lower case the value in each usage.

Like:

```
  <sch:let name="mainLanguage" value="if (normalize-space(//*[name(.)='gmd:MD_Metadata' or @gco:isoType='gmd:MD_Metadata']/gmd:language/gmd:LanguageCode/@codeListValue) != '')
                                       then lower-case(normalize-space(//*[name(.)='gmd:MD_Metadata' or @gco:isoType='gmd:MD_Metadata']/gmd:language/gmd:LanguageCode/@codeListValue))
                                       else if (contains(//*[name(.)='gmd:MD_Metadata' or @gco:isoType='gmd:MD_Metadata']/gmd:language/gco:CharacterString,';'))
                                            then lower-case(normalize-space(substring-before(//*[name(.)='gmd:MD_Metadata' or @gco:isoType='gmd:MD_Metadata']/gmd:language/gco:CharacterString,';')))
                                            else lower-case(//*[name(.)='gmd:MD_Metadata' or @gco:isoType='gmd:MD_Metadata']/gmd:language/gco:CharacterString)"/>
```

jodygarnett:
Applied @josegar74 feedback, and rebased from 3.10.x branch

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-non-multilingual.sch
ianwallen: @jodygarnett - Looks good.  Thanks you for applying the fix.

Issue #88, opened 2020-09-11
Metadata tab missing from advanced view.


ianwallen:
Geonetwork core has a metadata tab in the advanced view. But this does not exists in hnap
![image](https://user-images.githubusercontent.com/1868233/92932149-dc441d80-f41a-11ea-97d5-01ec7e64d5b0.png)

Looking to see the file identifier and other metadata information

I would expect the metadata standard fields to also be displayed as non editable field.

Odd that there is a language under identification and under metadata - not sure what the difference is?
I would also expect the metadata character set to be in metadata?  But I see it in identification tab - should probably investigate this as well.

ianwallen:
Tab was there but it was not displayed due to missing string values.

![image](https://user-images.githubusercontent.com/1868233/98668612-80b4e400-2326-11eb-8290-98ded85239b2.png)

It was fixed in the following commit.

https://github.com/metadata101/iso19139.ca.HNAP/commit/46995fca7637cad8d2620b963284c40394a94f88

PR #119, opened 2020-11-04
Added some strings from the iso19139 mostly to fix the xsd validation error messages

ianwallen:
Added some strings from the iso19139 mostly to fix the xsd validation error messages. Added some other missing strings as well.

Before this fix validation message may look like this.

image

After this fix the message should look something like this

image


jodygarnett (approved):
Thanks! I was about to go hunting for "missing" error messages with respect to schema validation.

ianwallen:
Also fixes #88

PR #123, opened 2020-11-09
adds reference to default iso19139 formatter

pvgenuchten:
fixes layout and schema.org annotations

in /srv/api/records/{uuid}

image


ianwallen:
This makes it better than what was there before.  But there are some issues with displaying the code list values.

image

image

And in French is looks like this.

image

Is this an easy fix?

josegar74:
I see this template that should retrieve the label:

https://github.com/geonetwork/core-geonetwork/blob/feef9919e0eaa40f9ef263fc5e9cedab5d31f3c8/schemas/iso19110/src/main/plugin/iso19110/formatter/xsl-view/view.xsl#L340-L364

Seem should be extracting the label value. In the screenshot seem it's displayed, but also the text value in the codelist element, needs to be review as the codelist in HNAP have always the text also, like:

```
<gmd:hierarchyLevel>
    <gmd:MD_ScopeCode codeList="http://nap.geogratis.gc.ca/metadata/register/napMetadataRegister.xml#IC_108" 
                      codeListValue="RI_622">dataset; jeuDonnées</gmd:MD_ScopeCode>
  </gmd:hierarchyLevel>
```

The problem seem this line:

https://github.com/geonetwork/core-geonetwork/blob/3fd0c87214fce344b4e91933efe5a3a9a6f29da4/schemas/iso19139/src/main/plugin/iso19139/formatter/xsl-view/view.xsl#L376

Going to apply a fix with the changes:

fix-codelists

josegar74:
@ianwallen please check with the updates and if all looks fine, please merge it.

ianwallen (approved):

PR #125, opened 2020-11-10
Add /gmd:MD_Metadata/gmd:language and  gmd:characterSet to the basic view in editor

ianwallen:
Add /gmd:MD_Metadata/gmd:language and  gmd:characterSet to the basic view in editor page as these fields are part of HNAP.

Users should not be changing language so made the gmd:language read only for the basic editor page but it can be modified on the advanced view as it was before this change.

Here is a sample of what it looks like.

image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
@@ -310,7 +310,7 @@
         <section name="gmd:MD_Metadata">
           <field xpath="/gmd:MD_Metadata/gmd:fileIdentifier"/>
           <field xpath="/gmd:MD_Metadata/gmd:dateStamp"/>
-          <field xpath="/gmd:MD_Metadata/gmd:language" or="language" in="/gmd:MD_Metadata"/>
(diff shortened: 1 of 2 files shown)


src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
```diff
@@ -310,6 +310,8 @@
         <section name="gmd:MD_Metadata">
           <field xpath="/gmd:MD_Metadata/gmd:fileIdentifier"/>
           <field xpath="/gmd:MD_Metadata/gmd:dateStamp"/>
+          <field xpath="/gmd:MD_Metadata/gmd:language" or="language" in="/gmd:MD_Metadata"/>
```
josegar74: `or` and `in` attributes require also the element to be defined in the `flatModeExceptions` list, so the editor adds the `+` button if the element if doesn't exist in the metadata. 

But in this case as in the default view is a readonly element I don't think it's required.

ianwallen: Fixed - thank you for the review.

josegar74 (approved):

PR #131, opened 2020-11-16
Updated hnap template to mean HNAP level 1 field requirements.

ianwallen:
Updated hnap template to mean HNAP level 1 field requirements.
Also fixed some bugs with the core subject thesaurus.


ianwallen:
Approving as this is mostly a DFO template.

Issue #107, opened 2020-10-27
POM issues with duplicate dependency and odd test resources


ianwallen:
Anyone know why we have the same dependency artifact twice in the pom.xml?

```
 <dependency>
     <groupId>${project.groupId}</groupId>
     <artifactId>core</artifactId>
     <version>${gn.project.version}</version>
    </dependency>
    <dependency>
      <groupId>${project.groupId}</groupId>
      <artifactId>schema-core</artifactId>
      <version>${project.version}</version>
      <type>test-jar</type>
      <scope>test</scope>
    </dependency>
    <dependency>

```
Also anyone know why the plugin folder is marked as a test resource and why there is reference to a test resource folder that does not exists?

```
<testResources>
      <testResource>
        <directory>src/test/resources/org/fao/geonet</directory>
      </testResource>
      <testResource>
        <directory>src/main/plugin/iso19139.ca.HNAP</directory>
      </testResource>
    </testResources>
```

I noticed that these same issues exist in schema iso19139 - So may be they were copied over from there?

jodygarnett:
Q: Anyone know why we have the same dependency artifact twice in the pom.xml?
A: Those are two different things:

* `core` references https://github.com/geonetwork/core-geonetwork/tree/master/core
* `schema-core` references https://github.com/geonetwork/core-geonetwork/tree/master/schemas/schema-core

`schema-core` publishes a couple different artifacts:

* `jar`used to help write schema plugins 
* `test-jar`often used for resources or base classes supporting the writing of test cases

jodygarnett:
Q: Also anyone know why the plugin folder is marked as a test resource and why there is reference to a test resource folder that does not exists?
A: This is indeed hard to figure out and I am not sure I have it right

We need the plugin `iso19139.ca.HNAP` content available to run tests against it:

* It is packaged for other modules as a `zip`, ... web app for example can depend on the `zip` and include in `war`
* But before we package we would like to test the contents, `testResource` is how to do this ...

I believe `src/test/resources` is just the normal `testResource` contents included to aid in unit testing

jodygarnett:
I guess a fix for this would be to better document how schema plugins are written and tested.

ianwallen:
My eyes were playing games with me - I did not notice core vs schema-core.  My mistake ignore that issue.

Regarding the testResource - I believe some plugins have this and some don't. i.e. iso19115-3.2018, dublin-core.  
[...]

PR #136, opened 2020-11-23
Consider plugin folder a resource, exclude contents from maven-jar-plugin

jodygarnett:
Setting up a build change from this discussion https://github.com/metadata101/iso19139.ca.HNAP/issues/107 changing the plugin folder to be a resource, and making sure the contents are excluded from maven-jar-plugin (since they are being packaged as a zip).

Traditionally schema plugins consider the plugin folder as a testResource, which is not quite accurate.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -63,6 +63,7 @@
 
   <build>
     <resources>
+      <!-- packaged into jar -->
       <resource>
         <directory>src/main/config/codelist/local/thesauri/theme</directory>
         <targetPath>thesarus/theme</targetPath>
@@ -82,18 +83,31 @@
       <resource>
         <directory>src/main/resources</directory>
       </resource>
+      <!-- packaged into zip -->
+      <resource>
+        <directory>src/main/plugin/iso19139.ca.HNAP</directory>
+      </resource>
     </resources>
-
     <testResources>
       <testResource>
         <directory>src/test/resources/org/fao/geonet</directory>
       </testResource>
-      <testResource>
-        <directory>src/main/plugin/iso19139.ca.HNAP</directory>
-      </testResource>
     </testResources>
 
     <plugins>
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <includes>
+            <include>ca/**</include>
+            <include>META-INF/**</include>
+            <include>thesarus/**</include>
+            <include>xsl/**</include>
+            <include>config-spring-geonetwork.xml</include>
+            <include>rebel.xml</include>
+          </includes>
+        </configuration>
+      </plugin>
       <plugin>
         <artifactId>maven-assembly-plugin</artifactId>
         <executions>


pom.xml
```diff
@@ -82,18 +83,40 @@
...
+            <exclude>sample-data/**</exclude>
+            <exclude>schema/**</exclude>
+            <exclude>schematron/**</exclude>
+            <exclude>templates/**</exclude>
+            <exclude>*.xsl</exclude>
+            <exclude>*.xml</exclude>
+            <exclude>*.xsd</exclude>
+          </excludes>
+        </configuration>
+      </plugin>
```
ianwallen, line 110: Instead of attempting to exclude all the file from the HNAP plugin (which could change and make it difficult to keep this up to date) would the following work when including the resource.

```
      <!-- packaged into zip -->
      <resource>
        <directory>src/main/plugin/iso19139.ca.HNAP</directory>
        <!-- Exclude from JAR -->
        <excludes>
          <exclude>**</exclude>
        </excludes>
      </resource>
```

I did a test locally and it seems to make the same jar and zip file. (I did not do a full compare of the zips to ensure they were the same)
Also IntelliJ does not seem to look at the exclude so the folder looks like a normal resource folder.

image

Note that src/test/resources/org/fao/geonet does not appear as a test resource even though it is identified above - that folder does not exists and I'm not sure if it should be removed?

I'm not a maven export so this may cause other issues that I'm not aware.

jodygarnett: I had just come to the same conclusion myself, now using includes to only grab the normal jar content.

And yes we expect this to look like a normal resource folder for IntelliJ (since it is a normal resource folder - it is just that we package this thing into a `jar` and `zip` for distribution.

aside: by the time the jar plugin sees the content it has all been flattened down into a single folder structure combining the built classes, and all the resource folders...

ianwallen: @jodygarnett 
I looked into the include as well and that should work but it still requires a change when a new folder is added. I find it easier to simply remove the `<plugin>` code altogether and use the suggested code that I included which includes the resource so that it will be in the zip(and show up properly in intellij). But then excludes all files so that it will not be in the jar.  I think this would be cleaner and will not cause issues when new data is added.

jodygarnett: @ianwallen I did not quite understand your code, since the assembly for the zip already has that logic:
``` 
<fileSets>
    <fileSet>
      <directory>src/main/plugin/</directory>
      <outputDirectory></outputDirectory>
      <useDefaultExcludes>true</useDefaultExcludes>
    </fileSet>
  </fileSets>
```

So what I am missing from your approach is how to ensure the tests can see the plugin contents?

Note if the plugin folder is not used by the test cases we can solve this very quickly, by removing `src/main/plugin/iso19139.ca.HNAP` folder as resource and as a testResource. 

It will still be packaged in the assembly above.

ianwallen: @jodygarnett 
I knew there was something basic that I was missing.  Locally I don't run the tests. 

Changes look good.

ianwallen (approved):

Issue #130, opened 2020-11-13
record formatters/xml not returning record that is valid.


ianwallen:
Steps to reproduce.

Create a metadata record.  (can use sample metadata records)
Ensure that it is valid

Use the following api to get the xml record.
/{portal}/api/0.1/records/{metadataUuid}/formatters/xml
http://localhost/geonetwork/doc/api/index.html#/records/getRecordAsXmlOrJSON_1

Now attempt to load and validate this record with FGP

In our example - when we attempted to load the record to FGP we got the following errors.
![image](https://user-images.githubusercontent.com/1868233/99081909-d8a75100-2599-11eb-9ea2-5b6986ac3c5b.png)

There seems to be some issues with the export  - I see that the original data that passes validation contained the following. (when editing the metadata record in xml mode)
```
<gmd:distributionFormat>
            <gmd:MD_Format>
               <gmd:name>
                  <gco:CharacterString>DOC</gco:CharacterString>
               </gmd:name>
               <gmd:version>
                  <gco:CharacterString>1.32</gco:CharacterString>
               </gmd:version>
            </gmd:MD_Format>
         </gmd:distributionFormat>
```
 
And the one that was generated by the API looks like the following.

```
<gmd:distributionFormat>
          <gmd:MD_Format>
            <gmd:name xsi:type="gmd:PT_FreeText_PropertyType">
              <gco:CharacterString>DOC</gco:CharacterString>
              <gmd:PT_FreeText/>
            </gmd:name>
            <gmd:version>
              <gco:CharacterString>1.32</gco:CharacterString>
            </gmd:version>
          </gmd:MD_Format>
        </gmd:distributionFormat>
```

It is putting the gmd:PT_FreeText in the gmd:name and it should not

And it is also moving the <gmd:locale> was moved to the top before <gmd:hierarchyLevel> - not sure what that seems to be causing an error?

ianwallen:
I can get similar errors if I do the following quicker steps.

1 - Updated settings to remove the schema location when validating

![image](https://user-images.githubusercontent.com/1868233/99087307-ff1cba80-25a0-11eb-8cdf-a21eb5b18c1c.png)

2 - load sample data (which should include "Harbour Porpoise Presence, Maritimes Region")
3 - export the sample record as a zip
4 - attempt to import the sample record with validation.

![image](https://user-images.githubusercontent.com/1868233/99087763-85d19780-25a1-11eb-92f9-3e65678d5ea0.png)

[...]

PR #137, opened 2020-11-23
Fix locale placement when saving data. 

ianwallen:
It was preventing exported records from being imported.


ianwallen:
@josegar74 
This is continuance of the following issue.
https://github.com/metadata101/iso19139.ca.HNAP/issues/130

You had fixed the distribution format issue with PR https://github.com/metadata101/iso19139.ca.HNAP/pull/133

This PR is to fix the gmd:locale issue.

PR #138, opened 2020-11-24
Add french translation text to HNAP schema

wangf1122:


wangf1122:
There will be more translation from other module that need to be committed in this pull request. Will reopen once it's done.

PR #139, opened 2020-11-26
Add French (multilingual) capacity to MD_TopicCategoryCode codelist with value content. Add more French labels.

wangf1122:


src/main/plugin/iso19139.ca.HNAP/loc/fre/strings.xml
```diff
@@ -60,8 +60,8 @@
   <xsdErrorIs>L'erreur est : </xsdErrorIs>
   <inElement> contenu dans </inElement>
   <enum1>La valeur </enum1>
-  <enum2> n'est pas valide pour le champ </enum2>
-  <enum3>Les valeurs possibles sont : </enum3>
+  <enum2> n'est pas valide pour le champ</enum2>
```
ianwallen, line 63: If you remove the space at the end, I believe this will cause an issue with the field that is appended after the text?

ianwallen (commented):
Please add a description explaining what these changes will correct.

Could you explain why codelist value is added instead of using code label?

I noticed that you are changing `<enum*>` values.  Is this PR dependent on PR
https://github.com/geonetwork/core-geonetwork/pull/5220. If so the then is should be identified in the description.

src/main/plugin/iso19139.ca.HNAP/loc/fre/strings.xml
wangf1122: No. I am fixing an issue instead. The xsl calling enum will add extra space. It will be double spaced with this space as well. So I have to remove it.

wangf1122:
> Please add a description explaining what these changes will correct.
> 
> Could you explain why codelist value is added instead of using code label?
> 
> I noticed that you are changing `<enum*>` values. Is this PR dependent on PR
> [geonetwork/core-geonetwork#5220](https://github.com/geonetwork/core-geonetwork/pull/5220). If so the then is should be identified in the description.

Label is general description. For example:

entry>
      code>boundaries</code
	  value>Limites</value
      label>Limites politiques et administratives</label
      description>Exemples : limites de pays, de provinces, de départements, de communes</description
    </entry

you can see the label is not exact language translation text of the code. I tends to remain stuff there instead of changing it. So I decide to add new value in. So it won't affect other features.

wangf1122:
PR cancelled due to the incorrect fix.

PR #146, opened 2020-12-02
Remove help tag (comment out) from /fre/label.xml to make it identical to eng/labels.xml element structure.

wangf1122:
There are two issues.

1) One is the help tool tip box has some duplicated text due to same text existed from description and help.

image

The result:
image

2) Another issue is 99% of eng/labels contains only description without help. However fre/labels.xml contains a lot of help information which is inconsistent to the eng/label.xml (see one of the element in the following) and the help tool tip is a lot complicated.

image

This version removes/comment out all unwanted help information from fre/labels/xml.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/loc/eng/labels.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/loc/eng/labels.xml
@@ -387,7 +387,7 @@
     <element name="gmd:MD_ReferenceSystem" id="186.0">
         <label>Reference system</label>
         <description>Information about the reference system.</description>
-        <help>information about the reference system</help>
+        <!--<help>information about the reference system</help>-->
     </element>
     <element name="gmd:MD_RepresentativeFraction" id="56.0">
         <label>Representative fraction</label>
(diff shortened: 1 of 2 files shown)


src/main/plugin/iso19139.ca.HNAP/loc/eng/labels.xml
```diff
@@ -387,7 +387,7 @@
     <element name="gmd:MD_ReferenceSystem" id="186.0">
         <label>Reference system</label>
         <description>Information about the reference system.</description>
-        <help>information about the reference system</help>
+        <!--<help>information about the reference system</help>-->
```
jodygarnett: It may be cleaner to just remove these

wangf1122: > It may be cleaner to just remove these

I totally agree. I was just worrying if other people will object to "removing" it because they want to keep some information. I will wait a little bit and see if there is no other objection, then I will commit another version to have all of them removed.

ianwallen: @wangf1122  - I also recommend that you just remove the text.  If ever we need it again - it can be retrieved from git

wangf1122: Another version is committed. All comments are removed.

josegar74 (approved):

PR #164, opened 2021-01-08
Remove lower case conversion.

wangf1122:
The indexing lowercase conversion has brought issues to search API. In the current codelist of gmd:hierarchyLevel/gmd:MD_ScopeCode element it consists of upper cases values, 

image

However when we index these fields, we force them to be lower cases only. Therefore, the search engine cannot find related label from codelist.xml and its translator. Since no label found, the value will be the key itself.

image


ianwallen (approved):

ianwallen:
@josegar74 do you know why this would have been lowercased in the past?
Looking at the FGP/ECCC code repo - I can see that on April 15 2019 you made the change (d0aa1e48f834b4e4857d485d95f69a4b1c5a79b0) and the comment indicates
"Fix indexing of metadata type"

Just want to make sure that reverting this change will not introduce other issues.

PR #165, opened 2021-01-13
Replace supported file format to the thesaurus by the FGP version.

wangf1122:
@ianwallen 

I compare the supported file format file from FGP to HNAP one by one and found these gaps. Some of them are case sensitive issue some of them are just new file format. Do you think this is the proper fix to resolve the validation error? Please share your opinion.

### DIFF
--- a/src/main/config/codelist/local/thesauri/theme/GC_Resource_Formats.rdf
+++ b/src/main/config/codelist/local/thesauri/theme/GC_Resource_Formats.rdf
@@ -1244,4 +1244,196 @@
         <ns2:prefLabel xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en"
         >MapInfo</ns2:prefLabel>
     </rdf:Description>
+
+    <rdf:Description rdf:about="http://geonetwork-opensource.org/EC/resourceformat#BAG">
+    		<ns2:scopeNote xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">Definition</ns2:scopeNote>
+    		<ns2:scopeNote xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="fr">Definition</ns2:scopeNote>
+    		<ns2:prefLabel xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="fr">BAG</ns2:prefLabel>
+    		<rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
+    		<ns2:prefLabel xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">BAG</ns2:prefLabel>
+    </rdf:Description>
+
+    <rdf:Description rdf:about="http://geonetwork-opensource.org/EC/resourceformat#JP2">
+    		<ns2:scopeNote xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">Definition</ns2:scopeNote>
+    		<ns2:scopeNote xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="fr">Definition</ns2:scopeNote>
+    		<ns2:prefLabel xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="fr">JPEG 2000 (JP2)</ns2:prefLabel>
+    		<rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
+    		<ns2:prefLabel xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">JPEG 2000 (JP2)</ns2:prefLabel>
+    </rdf:Description>
+
+    <rdf:Description rdf:about="http://geonetwork-opensource.org/EC/resourceformat#JSONL">
+    		<ns2:scopeNote xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">Definition</ns2:scopeNote>
+    		<ns2:scopeNote xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="fr">Definition</ns2:scopeNote>
+    		<ns2:prefLabel xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="fr">JSON Lignes (JSONL)</ns2:prefLabel>
+    		<rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
+    		<ns2:prefLabel xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">JSON Lines (JSONL)</ns2:prefLabel>
+    </rdf:Description>
+
+    <rdf:Description rdf:about="http://geonetwork-opensource.org/EC/resourceformat#PPTX">
+    		<ns2:scopeNote xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">Definition</ns2:scopeNote>
+    		<ns2:scopeNote xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="fr">Definition</ns2:scopeNote>
+    		<ns2:prefLabel xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="fr">PPTX</ns2:prefLabel>
+    		<rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
+    		<ns2:prefLabel xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">PPTX</ns2:prefLabel>
+    </rdf:Description>
+
+    <rdf:Description rdf:about="http://geonetwork-opensource.org/EC/resourceformat#RTF">
+    		<ns2:scopeNote xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">Definition</ns2:scopeNote>
+    		<ns2:scopeNote xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="fr">Definition</ns2:scopeNote>
+    		<ns2:prefLabel xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="fr">RTF</ns2:prefLabel>
+    		<rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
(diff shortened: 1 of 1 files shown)


ianwallen:
@wangf1122 if these changes come from FGP then we may want to look at merging their exact copy into ours.  I noticed that you added all the changes at the end of the file.  
I suspect that there will be a requirement to sync this in the future again. So it would be easier if they are as similar as possible.

@geothorne do you know where the official thesaurus for this exists? Is the FGP the official at this time?  We should try to ensure that we are all using the same list so that we are able to import FGP data as well as be able to push data to FGP.  It would be nice if we can use an exact copy of the file.

wangf1122:
> @wangf1122 if these changes come from FGP then we may want to look at merging their exact copy into ours. I noticed that you added all the changes at the end of the file.
> I suspect that there will be a requirement to sync this in the future again. So it would be easier if they are as similar as possible.
> 
> @geothorne do you know where the official thesaurus for this exists? Is the FGP the official at this time? We should try to ensure that we are all using the same list so that we are able to import FGP data as well as be able to push data to FGP. It would be nice if we can use an exact copy of the file.

@ianwallen  I also notice the FGP is lack of some name space for which it's different from our version. The namespace is "xmlns:ns2="http://www.w3.org/2004/02/skos/core#". I don't know if it matters or not.

akhamdo:
@ianwallen 
This is the Thesaurus we received from FGP (@AndrewWSHenry  and @bo-lu )
[EC_Resource_Formats.zip](https://github.com/metadata101/iso19139.ca.HNAP/files/5815718/EC_Resource_Formats.zip)

wangf1122:
> @ianwallen
> This is the Thesaurus we received from FGP (@AndrewWSHenry and @bo-lu )
> [EC_Resource_Formats.zip](https://github.com/metadata101/iso19139.ca.HNAP/files/5815718/EC_Resource_Formats.zip)

So guess I will need to find a tool to do the merge? I notice their namespace are different from ours so the merge tool will be pretty confused. I will ignore all the namespace diff and just merge file formats with our own name space into the file.

wangf1122:
@ianwallen  I did another round of merge. There are still a lot of manual process needed as the two files contains too many diffs. I did one way merge from FGP -> HNAP. So only added more stuff and not deleting any existing stuff.

ianwallen:
@wangf1122 
You missed 
MapInfo (TAB)

Also to make it cleaner - we can remove the namespace and define it at the top.

Once I updated that and did all the compares I see the following

### JPEG
image
image

Left is FGP and right it ours.
So it looks like there 2 JPEG2 and JPEG 2000
Looks at the specification, JPEG2 and JPEG 2000 seem to be the same.  So it looks like we should remove JPEG2 and JPEG 2000 and use the FGP JP2.

### MapInfo
image
image

Is there are difference between "Mapinfo" and "MapInfo (TAB)"? Not sure which one is correct? 
This could be another case where FGP merged them into one as they are the same?

### Other possible changes
Looks like 
```
	<rdf:Description rdf:about="http://geonetwork-opensource.org/EC/resourceformat#IOS">
		<ns2:scopeNote xml:lang="en">Definition</ns2:scopeNote>
		<ns2:scopeNote xml:lang="fr">Definition</ns2:scopeNote>
		<ns2:prefLabel xml:lang="fr">IOS</ns2:prefLabel>
		<rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
		<ns2:prefLabel xml:lang="en">IOS</ns2:prefLabel>
	</rdf:Description>
```
Changed to 
```
	<rdf:Description rdf:about="http://geonetwork-opensource.org/EC/resourceformat#IPA">
		<ns2:scopeNote xml:lang="en">Definition</ns2:scopeNote>
		<ns2:scopeNote xml:lang="fr">Definition</ns2:scopeNote>
		<ns2:prefLabel xml:lang="fr">Application iOS (IPA)</ns2:prefLabel>
		<rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
		<ns2:prefLabel xml:lang="en">iOS Application (IPA)</ns2:prefLabel>
	</rdf:Description>
```

and 

```
	<rdf:Description rdf:about="http://geonetwork-opensource.org/EC/resourceformat#JSON-LD">
		<ns2:scopeNote xml:lang="en">Definition</ns2:scopeNote>
		<ns2:scopeNote xml:lang="fr">Definition</ns2:scopeNote>
		<ns2:prefLabel xml:lang="fr">JSON-LD</ns2:prefLabel>
		<rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
		<ns2:prefLabel xml:lang="en">JSON-LD</ns2:prefLabel>
	</rdf:Description>
```
Changed to 

```
	<rdf:Description rdf:about="http://geonetwork-opensource.org/EC/resourceformat#JSONLD">
		<ns2:scopeNote xml:lang="en">Definition</ns2:scopeNote>
		<ns2:scopeNote xml:lang="fr">Definition</ns2:scopeNote>
		<ns2:prefLabel xml:lang="fr">JSON-LD</ns2:prefLabel>
		<rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
		<ns2:prefLabel xml:lang="en">JSON-LD</ns2:prefLabel>
	</rdf:Description>
```
... There are lots of these.

### Extra new items that we have but don't exists in FGP

I could not find anything on these - not sure if they were just removed?
```
	<rdf:Description rdf:about="http://geonetwork-opensource.org/EC/resourceformat#AI">
		<ns2:scopeNote xml:lang="en">Definition</ns2:scopeNote>
		<ns2:scopeNote xml:lang="fr">Definition</ns2:scopeNote>
		<ns2:prefLabel xml:lang="fr">AI</ns2:prefLabel>
		<rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
		<ns2:prefLabel xml:lang="en">AI</ns2:prefLabel>
	</rdf:Description>
	<rdf:Description rdf:about="http://geonetwork-opensource.org/EC/resourceformat#AMF">
		<ns2:scopeNote xml:lang="en">Definition</ns2:scopeNote>
		<ns2:scopeNote xml:lang="fr">Definition</ns2:scopeNote>
		<ns2:prefLabel xml:lang="fr">AMF</ns2:prefLabel>
		<rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
		<ns2:prefLabel xml:lang="en">AMF</ns2:prefLabel>
	</rdf:Description>
```

There may be others - I did not check them all.

## Conclusion

In the end, if we have more items than FGP, then it will cause issue when we attempt to push data to FGP and if FGP has a different list than we have then we will have issues importing their data.

So I vote to make them the same.

The only thing that is different in our files that we should keep, is the `<ConceptScheme>`  because that is used to show multilingual tiltes for this thesaurus.  I think FGP should look into adding this to their file.

image

Here is the file with namespace cleaned up but still has all our extra items.  It was used for the above comparison.

[GC_Resource_Formats.zip](https://github.com/metadata101/iso19139.ca.HNAP/files/5816463/GC_Resource_Formats.zip)

(@AndrewWSHenry and @bo-lu ) - could you confirm if FGP has already gone through a cleanup exercise of this list?

@akhamdo and @geothorne -  Do we want to use the supplied FGP thesaurus as is and remove our extras so that we have the same list? - I recommend that we do.

ianwallen:
@jodygarnett could you also review the discussion on this issue so that this does not affect any of your clients.

wangf1122:
Thanks Ian. It look a lot cleaner. I thought we just do FGP->HNAP one way. Didn't know the two way thing (HNAP meta needs to be imported to FGP as well). Yeah so in this case, we'd better to make it unique. So we don't have anything extra nor FGP. So it's N to N mapping relationship.

ianwallen:
@wangf1122 it seems that we have decided to use the FGP version.
Could you update the PR so that it uses the FGP version with the exception of the `<ConceptScheme>` which needs to be added to the top

wangf1122:
> @wangf1122 it seems that we have decided to use the FGP version.
> Could you update the PR so that it uses the FGP version with the exception of the `<ConceptScheme>` which needs to be added to the top

Yes. I am testing locally and will push soon. I will also update the bug description.

wangf1122:
> > @wangf1122 it seems that we have decided to use the FGP version.
> > Could you update the PR so that it uses the FGP version with the exception of the `<ConceptScheme>` which needs to be added to the top
> 
> Yes. I am testing locally and will push soon. I will also update the bug description.

Hi @ianwallen ,

I totally nuked it and replaced by FGP version. The ConceptScheme remains. Also inserted FGP namespaces into the title tag definition.  See the screenshot.

image

src/main/config/codelist/local/thesauri/theme/GC_Resource_Formats.rdf
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+         xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
          xmlns:skos="http://www.w3.org/2004/02/skos/core#"
```
ianwallen: skos and ns2 are the same namespace. 
As FGP use ns2 - I suggest that you remove skos namespace and replace all occurrences with ns2. i.e. change `<skos:ConceptScheme>` to `<ns2:ConceptScheme>`

src/main/config/codelist/local/thesauri/theme/GC_Resource_Formats.rdf
```diff
@@ -1,7 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+         xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
```
ianwallen, line 3: This seems like a type.  I don't see **rdfs** used below but I do see **rdf**.  I think you need to drop the **s**.

ianwallen (commented):
There seems to be a couple namespace bugs to fix. But the rest looks good

src/main/config/codelist/local/thesauri/theme/GC_Resource_Formats.rdf
ianwallen: Sorry just saw the rdf namespace above.   I still don't see where rdfs is being used.  I guess we can keep it.

wangf1122: Yeah. The RDFs is from FGP as copy pasted. It will probably ok to remain. I will go replace the SKOs by NS2

ianwallen (approved):

PR #181, opened 2021-02-05
Fix multilingual thesaurus title grouping in keywords indexing

josegar74:
In the metadata detail page keywords from Core Subject Thesaurus are displayed twice:

keywords-1

This is caused by this code:

https://github.com/metadata101/iso19139.ca.HNAP/blob/c8901f36f968cdc7733258d08af45117d084d1ac/src/main/plugin/iso19139.ca.HNAP/index-fields/default.xsl#L334-L335

The code processes `gco:CharacterString` getting the correct title and `gmd:PT_FreeText` getting empty string (similar issue in the file `language-default.xsl`

With the change:

keywords-2


jodygarnett:
Thanks @josegar74 testing now.

jodygarnett (approved):
I tested locally using war overlay successfully.

PR #188, opened 2021-02-26
Change version to 2.3.1 to reflect hnap schema

### LINKED ISSUE #185: Version issue when attempting to use newer HNAP
I got the following error 

2021-02-18 08:59:17,233 ERROR [geonetwork.schemamanager] - Schema iso19139.ca.HNAP requires min Geonetwork version: 3.10.5, current is: 3.10-SNAPSHOT. Skip load schema.

And seem to be caused by the following lines.

https://github.com/metadata101/iso19139.ca.HNAP/blob/a1f26d73d543724e040b9f462fd83b4f56d5755a/src/main/plugin/iso19139.ca.HNAP/schema-ident.xml#L30-L31

If I change these to `3.10` then it seems to work.  
I also tried `3.10.6` - but it die not work? - Got the following error.

2021-02-18 09:34:17,906 ERROR [geonetwork.schemamanager] - Schema iso19139.ca.HNAP requires min Geonetwork version: 3.10.6, current is: 3.10-SNAPSHOT. Skip load schema.

@jodygarnett what should these values be set to? Should it be 3.10?  And if so how do we indicate that release 3.10.6 may not work on 3.10.4?

In either case 3.10.5 seems to be incorrect and should be modified.

jodygarnett:
Also set the appMinorVersionSupported to 3.10-SNAPSHOT (remember to change during release process)

This resolves https://github.com/metadata101/iso19139.ca.HNAP/issues/185

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schema-ident.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/schema-ident.xml
@@ -27,8 +27,9 @@
         xsi:schemaLocation="http://geonetwork-opensource.org/schemas/schema-ident http://geonetwork-opensource.org/schemas/schema-ident/schema-ident.xsd">
 	<name>iso19139.ca.HNAP</name>
 	<id>3693b080-4f89-11e4-916c-0800200c9a66</id>
-	<version>3.10.5</version>
-	<appMinorVersionSupported>3.10.5</appMinorVersionSupported>
+	<!-- hnap schema version -->
+	<version>2.3.1</version>
+	<appMinorVersionSupported>3.10-SNAPSHOT</appMinorVersionSupported>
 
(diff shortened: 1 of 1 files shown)


juanluisrp:
I'm wondering if this doesn't need to be updated too:
https://github.com/metadata101/iso19139.ca.HNAP/blob/a1f26d73d543724e040b9f462fd83b4f56d5755a/src/main/plugin/iso19139.ca.HNAP/schema-ident.xml#L30

juanluisrp (commented):
LGTM.

juanluisrp:
Merging this after geonetwork/core-geonetwork/pull/5452 (for `master`) and geonetwork/core-geonetwork/pull/5451 (for `3.10.x` has been merged)

PR #194, opened 2021-03-16
Schematron rules xpath fixes, and fixes in xpath in config-editor.xsl…

wangf1122:
@ianwallen 

These three validation error are not showing. I found some workaround just want to be sure if it's ok.

image

1) Some adjustment made to gmd:EX_TemporalExtent xpath in the schematron-rules-common.sch. But I am so sure of the change in config-editor.xml. I have to remove templateModeOnly="true"

image

Otherwise I can never got the validation error target to the right field. Here is the result and you will see "Temporal Event" is added with tooltip and the error is attached to this field.
image

2) I found the keyword error is not showing in both basic and advanced mode. So I basically change the xpath from //gmd:identificationInfo/*/gmd:descriptiveKeywords/gmd:MD_Keywords/gmd:keyword to //gmd:identificationInfo/*/gmd:descriptiveKeywords
So I could get it working. Not sure the science behind. Here is the result:
image

3) I also found some strings missing. So I added them, here is the fixed result:
image

Let me know if the fix are ok or not. Thanks.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
@@ -517,10 +517,10 @@
 
     <!-- Keywords -->
     <sch:rule context="//gmd:identificationInfo/*/gmd:descriptiveKeywords">
-      <sch:let name="missing" value="not(string(gco:CharacterString))
+      <sch:let name="missing" value="not(string(gmd:MD_Keywords/gmd:keyword[1]/gco:CharacterString))
             or (@gco:nilReason)" />
 
-      <sch:let name="missingOtherLang" value="not(string(gmd:PT_FreeText/gmd:textGroup/gmd:LocalisedCharacterString[@locale=concat('#', $altLanguageId)]))" />
+      <sch:let name="missingOtherLang" value="not(string(gmd:MD_Keywords/gmd:keyword[1]/gmd:PT_FreeText/gmd:textGroup/gmd:LocalisedCharacterString[@locale=concat('#', $altLanguageId)]))" />
 
       <sch:assert
         test="not($missing) and not($missingOtherLang)"
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
@@ -516,7 +516,7 @@
 
 
     <!-- Keywords -->
-    <sch:rule context="//gmd:identificationInfo/*/gmd:descriptiveKeywords/gmd:MD_Keywords/gmd:keyword">
+    <sch:rule context="//gmd:identificationInfo/*/gmd:descriptiveKeywords">
       <sch:let name="missing" value="not(string(gco:CharacterString))
             or (@gco:nilReason)" />
 
--- a/src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
@@ -468,7 +468,7 @@
                    in="/gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification"
                    or="resourceConstraints"/>
 
-          <field name="temporalRangeSection" templateModeOnly="true"
+          <field name="temporalRangeSection"
                  xpath="/gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification/gmd:extent/gmd:EX_Extent/gmd:temporalElement"
                  if="count(gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification) > 0">
             <template>
@@ -505,7 +505,7 @@
             </template>
           </field>
 
-          <field name="temporalRangeSection" templateModeOnly="true"
+          <field name="temporalRangeSection"
                  xpath="/gmd:MD_Metadata/gmd:identificationInfo/srv:SV_ServiceIdentification/gmd:extent/gmd:EX_Extent/gmd:temporalElement"
                  if="count(gmd:MD_Metadata/gmd:identificationInfo/srv:SV_ServiceIdentification) > 0">
             <template>
--- a/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-non-multilingual.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-non-multilingual.xml
@@ -38,7 +38,7 @@
   <ContactOrganisationName>Contact - Organization name is required</ContactOrganisationName>
   <CitedResponsiblePartyOrganisationName>Cited Responsible Party - Organization name is required</CitedResponsiblePartyOrganisationName>
   <DistributorOrganisationName>Distributor - Organization name is required</DistributorOrganisationName>
-  
+
   <ContactElectronicMail>Contact - Electronic mail address is required</ContactElectronicMail>
   <CitedResponsiblePartyElectronicMail>Cited Responsible Party - Electronic mail address is required</CitedResponsiblePartyElectronicMail>
   <DistributorElectronicMail>Distributor - Electronic mail address is required</DistributorElectronicMail>
@@ -50,4 +50,8 @@
   <ECThesaurusOrg>Thesaurus cited responsible party organisation is required</ECThesaurusOrg>
   <ECThesaurusRole>Thesaurus cited responsible party role is required</ECThesaurusRole>
   <ECThesaurusEmail>Thesaurus cited responsible party email is required</ECThesaurusEmail>
+
+  <ResourceDescriptionContentType>Online resource description content type is not valid. Valid values are: Web Service,Service Web,Dataset,Données,API,Application,Supporting Document,Document de soutien</ResourceDescriptionContentType>
+  <ResourceDescriptionFormat>Online resource description format is not valid. Valid values are:</ResourceDescriptionFormat>
+  <ResourceDescriptionLanguage>Online resource description language is not valid. Should be a comma separated values of ISO-LANG-3 codes</ResourceDescriptionLanguage>
 </strings>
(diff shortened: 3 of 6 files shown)


ianwallen:
@wangf1122 
    All changes except for the `templateModeOnly="true"` look good to me.  I'm not sure why removing that corrected the problem.

@josegar74 
    Could you help identify the reason why templateModeOnly="true" would be required.  I could not find much information on this field and why it would or would not be required.  Not sure if it is broken?

wangf1122:
> @wangf1122
> All changes except for the `templateModeOnly="true"` look good to me. I'm not sure why removing that corrected the problem.
> 
> @josegar74
> Could you help identify the reason why templateModeOnly="true" would be required. I could not find much information on this field and why it would or would not be required. Not sure if it is broken?

I remove this because this seems hide the "temporal extend" for which the validation error is attached to. So as result, you wont see error at all. 

This is an example, see the highlighted area for missing "temporal extend"  title and error.
image

davidblasby:
As far as I can tell, the templateModeOnly is not used in java.  I think this is where it gets set to;

https://github.com/geonetwork/core-geonetwork/blob/master/schemas/schema-core/src/main/java/org/fao/geonet/kernel/schema/editorconfig/Field.java#L239-L259

Here's its documentation;
```
- **templateModeOnly** (Optional) Fixed value: **true**

  Define if the template mode should be the only mode used. In that case, the field is always
  displayed based on the XML template snippet field configuration. Default is false.
```

It does appear to be processed in xslt;

https://github.com/geonetwork/core-geonetwork/blob/master/web/src/main/webapp/xslt/ui-metadata/form-configurator.xsl

wangf1122:
I attempt to make the validation rule to xPath of "//gmd:identificationInfo/*/gmd:extent/gmd:EX_Extent/gmd:temporalElement/gmd:EX_TemporalExtent/gmd:extent/gml:TimePeriod//gml:beginPosition". But still no luck with this templateModeOnly="true".

josegar74:
About the `templateModeOnly` it's used to display the template defined in the `config-editor.xml` for the related field, independently if the metadata xml contains it or not. 

I'm checking as not really sure why that affects the validation.

wangf1122:
> About the `templateModeOnly` it's used to display the template defined in the `config-editor.xml` for the related field, independently if the metadata xml contains it or not.
> 
> I'm checking as not really sure why that affects the validation.

FYI. Even I move the validation to the very low level of the begin date. The xPath is like: "//gmd:identificationInfo/*/gmd:extent/gmd:EX_Extent/gmd:temporalElement/gmd:EX_TemporalExtent/gmd:extent/gml:TimePeriod/*/gml:beginPosition". It still doesn't work without removing this templateModeOnly

josegar74:
The difference is not really related to the `templateModeOnly` attribute, but to the configuration of the fields in the template:

- **With** `templateModeOnly="true"` it uses the field template defined in `config-editor.xml`. The advantage of this mode is that it displays the elements independently if are defined or not in the xml. 

  But in this case the fields in the template use the angularjs directive`gn-date-picker` that seems to be the problem, this directive doesn't add to the elements the `ref` information that is added to other fields rendered without directives, so the inline validation messages doesn't work.


temporal-extent-directive


- **Without** `templateModeOnly="true"` (or set to `false`) it just renders the field content as usual, so if exists it's displayed, otherwise not. I don't get the usage of a field with a template defined without `templateModeOnly="true"`. The validation works as the fields are rendered using the standard way adding the `ref` elements.

  If `gmd:beginPosition` or `gmd:endPosition` would be configured on the top of the file similar to ` <for name="gco:Date" use="data-gn-date-picker"/>` would have the same issue.

temporal-extent-no-directive



If not really required the usage of the angularjs directive `gn-date-picker` for these fields. I would setup the fields without a template:

```
   <field name="temporalRangeSection"
                 xpath="/gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification/gmd:extent/gmd:EX_Extent/gmd:temporalElement"
                 if="count(gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification) > 0" />


  <field name="temporalRangeSection"
                 xpath="/gmd:MD_Metadata/gmd:identificationInfo/srv:SV_ServiceIdentification/gmd:extent/gmd:EX_Extent/gmd:temporalElement"
                 if="count(gmd:MD_Metadata/gmd:identificationInfo/srv:SV_ServiceIdentification) > 0" />
```


And use the `inflate-metadata.xsl` to add the `gmd:temporalElement` if missing (I think at least begin date is mandatory ?).

Using the standard rendering, I notice that the labels are not displayed and both fields are set as mandatory, something to review in the `labels.xml` file.

wangf1122:
> The difference is not really related to the `templateModeOnly` attribute, but to the configuration of the fields in the template:
> 
> * **With** `templateModeOnly="true"` it uses the field template defined in `config-editor.xml`. The advantage of this mode is that it displays the elements independently if are defined or not in the xml.
>   But in this case the fields in the template use the angularjs directive`gn-date-picker` that seems to be the problem, this directive doesn't add to the elements the `ref` information that is added to other fields rendered without directives, so the inline validation messages doesn't work.
> 
> temporal-extent-directive
> 
> * **Without** `templateModeOnly="true"` (or set to `false`) it just renders the field content as usual, so if exists it's displayed, otherwise not. I don't get the usage of a field with a template defined without `templateModeOnly="true"`. The validation works as the fields are rendered using the standard way adding the `ref` elements.
>   If `gmd:beginPosition` or `gmd:endPosition` would be configured on the top of the file similar to ` <for name="gco:Date" use="data-gn-date-picker"/>` would have the same issue.
> 
> temporal-extent-no-directive
> 
> If not really required the usage of the angularjs directive `gn-date-picker` for these fields. I would setup the fields without a template:
> 
> ```
>    <field name="temporalRangeSection"
>                  xpath="/gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification/gmd:extent/gmd:EX_Extent/gmd:temporalElement"
>                  if="count(gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification) > 0" />
> 
> 
>   <field name="temporalRangeSection"
>                  xpath="/gmd:MD_Metadata/gmd:identificationInfo/srv:SV_ServiceIdentification/gmd:extent/gmd:EX_Extent/gmd:temporalElement"
>                  if="count(gmd:MD_Metadata/gmd:identificationInfo/srv:SV_ServiceIdentification) > 0" />
> ```
> 
> And use the `inflate-metadata.xsl` to add the `gmd:temporalElement` if missing (I think at least begin date is mandatory ?).
> 
> Using the standard rendering, I notice that the labels are not displayed and both fields are set as mandatory, something to review in the `labels.xml` file.

I agree with your approach of "without". We don't need to add anything to the inflator as it already contains "gmd:temporalElement"
image

For mandatory field, this is a little confused me. Neither gml:beginPosition nor gml:endPosition is marked as mandatory in label, but they do appeared as mandatory.
image

ianwallen:
@wangf1122 

I guessing that since this is a special field the mandatory flag is in the template 

https://github.com/metadata101/iso19139.ca.HNAP/blob/35461610d2df9878024401b3bc80d1075e45c8a7/src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml#L477-L481

ianwallen:
On our local system I was just testing trying to remove the Temporal Extent and I cannot.  It keeps coming back.  Is this related to the template flag as well. Not sure if that is another bug?   I have not test this with the current changes from this branch.   If this turns out to be another bug then maybe we can create a separate issue for that.

wangf1122:
> On our local system I was just testing trying to remove the Temporal Extent and I cannot. It keeps coming back. Is this related to the template flag as well. Not sure if that is another bug? I have not test this with the current changes from this branch. If this turns out to be another bug then maybe we can create a separate issue for that.

I am good with separate issue created here

ianwallen:
Created the issue #200

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
```diff
@@ -516,11 +516,11 @@
 
 
     <!-- Keywords -->
-    <sch:rule context="//gmd:identificationInfo/*/gmd:descriptiveKeywords/gmd:MD_Keywords/gmd:keyword">
-      <sch:let name="missing" value="not(string(gco:CharacterString))
+    <sch:rule context="//gmd:identificationInfo/*/gmd:descriptiveKeywords">
+      <sch:let name="missing" value="not(string(gmd:MD_Keywords/gmd:keyword[1]/gco:CharacterString))
```
ianwallen, line 520: I don't believe this will work.

Previously the rule was for all //gmd:identificationInfo/*/gmd:descriptiveKeywords/gmd:MD_Keywords/gmd:keyword

Now the rule is only for the first gmd:keyword[1].

I'm guessing that it we were to add multiple keywords and the second+ keyword have issues. i.e. missing or does not contain other language, then the validation will still pass.  We will need to test this.

wangf1122: Previously the rule was set for the first keyword. Now we are pushing it back to "//gmd:identificationInfo/*/gmd:descriptiveKeywords" level. So it will be a list. In this case, I dont think it's useful to create a util method to integrator thru all of them because if the first element is missing, this keyword list will be empty anyway. So it will only needed to check the first keyword.

Edited: The reason we need to put keyword[1] is if you put two key words or more. For single one it's ok without.

ianwallen (changes requested):
Before this change the temporal extent looks like this.
image

After this change the temporal extents look like this.
image

The titles "Begin Date" and "End Date" are now missing and it now looks like they are both mandatory based on the "*" - but the validation seems to work correctly. 

Please look at putting back the titles - Also we need the old date selector which allowed entering only "year" for the date.

wangf1122:
> Before this change the temporal extent looks like this.
> image
> 
> After this change the temporal extents look like this.
> image
> 
> The titles "Begin Date" and "End Date" are now missing and it now looks like they are both mandatory based on the "*" - but the validation seems to work correctly.
> 
> Please look at putting back the titles - Also we need the old date selector which allowed entering only "year" for the date.

I did some workaround and had no luck putting the label back. The labels are clearly in the label.xml:

<element name="gml:beginPosition">
        <label>Begin Date</label>
        <description>Enter the date on which data collection began or is scheduled to begin. You may use the calendar to select the date or enter it in YYYY-MM-DD format for full date, YYYY-MM format for month/year or YYYY format for year.</description>
    </element>

And don't understand why it's not added. Here is two way we should go. 

1) We can revert the templateMode="true" back and deal with the validation.
2) We will keep looking for how to add the labels.

wangf1122:
The missing label also does appear in advanced mode. 
image

I created an issue for that:

https://github.com/metadata101/iso19139.ca.HNAP/issues/205

wangf1122:
This pull request is spited to two PR.

1. https://github.com/metadata101/iso19139.ca.HNAP/pull/206
2. https://github.com/metadata101/iso19139.ca.HNAP/pull/207

ianwallen:
I have merged the 2 related PR - you will have to correct the merge conflicts on the PR - or start a new PR.

wangf1122:
> I have merged the 2 related PR - you will have to correct the merge conflicts on the PR - or start a new PR.

This PR will get cancelled. New PR will be put in place.

PR #199, opened 2021-03-25
Updated security note so that it defaults to unknown when it is not defined

ianwallen:
Updated security note so that it defaults to unknown when it is not defined so that all document are assigned a security level according to ITSG-33

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/index-fields/language-default.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/index-fields/language-default.xsl
@@ -378,6 +378,8 @@
         </xsl:when>
         <xsl:otherwise>
           <Field name="secConstr" string="false" store="true" index="true"/>
+          <xsl:variable name="securityConstraintsUseLimitation" select="if ($isoLangId = 'fra') then 'Inconnu' else 'Unknown'"/>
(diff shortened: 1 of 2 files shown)


src/main/plugin/iso19139.ca.HNAP/index-fields/language-default.xsl
```diff
@@ -378,6 +378,8 @@
         </xsl:when>
         <xsl:otherwise>
           <Field name="secConstr" string="false" store="true" index="true"/>
+          <xsl:variable name="securityConstraintsUseLimitation" select="if ($isoLangId = 'fra') then 'Inconnu' else 'Unknown'"/>
+          <Field name="MD_SecurityConstraintsUseLimitation" string="{$securityConstraintsUseLimitation}" store="true" index="true"/>
```
jodygarnett, line 382: If undefined the security constraining use limitation level is shown as `Unknown` / `Inconnu`

jodygarnett: Discussion: Prior systems probably assumed Unclassified, this makes it more specific as Unknown (for example from external systems).

jodygarnett: Aside: This is only for the user interface / editor display, the actual XML record is not fixed.

Looking at a future PR to address this during import processing.

PR #202, opened 2021-04-09
Add default security constraint to FGP.19139.HNAP template.

wangf1122:
This is for FGP import template only. It will not impact other import jobs.

Security constraint is added if the xml metadata is lack of. Otherwise, leave as is. The overwrite existing security constraint to default code is added as well but it's been disabled. In the future, we can enable it if it's needed.

### DIFF
--- a/src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
+++ b/src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
@@ -126,17 +126,6 @@
     </xsl:copy>
   </xsl:template>
 
-  <!--Overwrite security constraint to default unclassified if presented in metadata xml-->
-  <!--<xsl:template match="gmd:resourceConstraints/gmd:MD_SecurityConstraints/gmd:classification">
-    <xsl:element name="gmd:classification">
-      <xsl:element name="gmd:MD_ClassificationCode">
-        <xsl:attribute name="codeList">http://nap.geogratis.gc.ca/metadata/register/napMetadataRegister.xml#IC_96</xsl:attribute>
-        <xsl:attribute name="codeListValue">RI_484</xsl:attribute>
-        <xsl:text>unclassified; nonClassifié</xsl:text>
-      </xsl:element>
-    </xsl:element>
-  </xsl:template>-->
-
   <xsl:template match="node()|@*">
     <xsl:copy>
       <xsl:apply-templates select="node()|@*"/>
--- a/src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
+++ b/src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
@@ -102,6 +102,41 @@
     </xsl:copy>
   </xsl:template>
 
+  <!--Add default unclassified as security constraint if missing from metadata xml-->
+  <xsl:template match="gmd:MD_DataIdentification">
+    <xsl:copy copy-namespaces="no">
+      <xsl:choose>
+        <xsl:when test="gmd:resourceConstraints/gmd:MD_SecurityConstraints/gmd:classification/gmd:MD_ClassificationCode">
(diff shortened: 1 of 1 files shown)


src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
```diff
@@ -102,6 +102,41 @@
...
+              </xsl:element>
+            </xsl:element>
+          </xsl:element>
+        </xsl:otherwise>
+      </xsl:choose>
+      <xsl:apply-templates select="node()|@*"/>
+    </xsl:copy>
+  </xsl:template>
+
+  <!--Overwrite security constraint to default unclassified if presented in metadata xml-->
```
ianwallen: Please remove comment as we are not overwriting existing classification values.  If it fails to validate user will have to change manually modify the value.

The only thing that we may overwrite is if the user note is using an old value then we may want to change it to the correct value.  Otherwise the user would have to change that as well.
See https://github.com/metadata101/iso19139.ca.HNAP/pull/195
i.e. "Protect A" could be changed to "Protected A"

src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
```diff
@@ -102,6 +102,41 @@
...
   </xsl:template>
 
+  <!--Add default unclassified as security constraint if missing from metadata xml-->
+  <xsl:template match="gmd:MD_DataIdentification">
+    <xsl:copy copy-namespaces="no">
+      <xsl:choose>
+        <xsl:when test="gmd:resourceConstraints/gmd:MD_SecurityConstraints/gmd:classification/gmd:MD_ClassificationCode">
+        </xsl:when>
+        <xsl:otherwise>
+          <xsl:element name="gmd:resourceConstraints">
```
ianwallen: I believe it would be cleaner to simply add the element instead of using xsl:element/xsl:attribute/xsl:text.
i.e. 
https://github.com/metadata101/iso19139.ca.HNAP/blob/16537ee38e8492d1fb5d971f150e0016357cb46e/src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl#L239-L250

src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
```diff
@@ -102,6 +102,41 @@
     </xsl:copy>
   </xsl:template>
 
+  <!--Add default unclassified as security constraint if missing from metadata xml-->
+  <xsl:template match="gmd:MD_DataIdentification">
+    <xsl:copy copy-namespaces="no">
+      <xsl:choose>
+        <xsl:when test="gmd:resourceConstraints/gmd:MD_SecurityConstraints/gmd:classification/gmd:MD_ClassificationCode">
```
ianwallen: May want the try something like the following
`<xsl:if test="not(gmd:resourceConstraints/gmd:MD_SecurityConstraints/gmd:classification/gmd:MD_ClassificationCode)">`
instead of creating an empty condition.

ianwallen (changes requested):

src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
wangf1122: Removed.

src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
wangf1122: Good suggestion. I will try.

wangf1122:
@ianwallen 

I made some clean up to the code. 

1) Removed the Choose and replaced with a simple "if" statement.
2) Put xml element directly without using <xsl:element>

ianwallen (approved):

PR #206, opened 2021-04-20
Keyword validation error message missing.

wangf1122:
After the fix:

image

This PR was spited from https://github.com/metadata101/iso19139.ca.HNAP/pull/194 as sub PR.


ianwallen (approved):
I find it odd that it only checks the first keyword to ensure that it is multilingual.  I tested and it seems to work however I believe it works because the update-fixe-info.xsl is updating the keyword list.

I will approve it as I cannot think of a better way.  And the change does produce the expected results of displaying the error message correclty.

Issue #210, opened 2021-04-30
Validation text not compatible with Temporal Extent when using templateModeOnly=true


wangf1122:
The difference is not really related to the `templateModeOnly` attribute, but to the configuration of the fields in the template:

- **With** `templateModeOnly="true"` it uses the field template defined in `config-editor.xml`. The advantage of this mode is that it displays the elements independently if are defined or not in the xml. 

  But in this case the fields in the template use the angularjs directive`gn-date-picker` that seems to be the problem, this directive doesn't add to the elements the `ref` information that is added to other fields rendered without directives, so the inline validation messages doesn't work.

![temporal-extent-directive](https://user-images.githubusercontent.com/1695003/112499748-16a65400-8d88-11eb-83f1-e97e41e68ce0.png)

- **Without** `templateModeOnly="true"` (or set to `false`) it just renders the field content as usual, so if exists it's displayed, otherwise not. I don't get the usage of a field with a template defined without `templateModeOnly="true"`. The validation works as the fields are rendered using the standard way adding the `ref` elements.

  If `gmd:beginPosition` or `gmd:endPosition` would be configured on the top of the file similar to ` <for name="gco:Date" use="data-gn-date-picker"/>` would have the same issue.

![temporal-extent-no-directive](https://user-images.githubusercontent.com/1695003/112499799-1dcd6200-8d88-11eb-9f62-db41f1b4dda8.png)

If not really required the usage of the angularjs directive `gn-date-picker` for these fields. I would setup the fields without a template:

```
   <field name="temporalRangeSection"
                 xpath="/gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification/gmd:extent/gmd:EX_Extent/gmd:temporalElement"
                 if="count(gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification) > 0" />


  <field name="temporalRangeSection"
                 xpath="/gmd:MD_Metadata/gmd:identificationInfo/srv:SV_ServiceIdentification/gmd:extent/gmd:EX_Extent/gmd:temporalElement"
                 if="count(gmd:MD_Metadata/gmd:identificationInfo/srv:SV_ServiceIdentification) > 0" />
```

And use the `inflate-metadata.xsl` to add the `gmd:temporalElement` if missing (I think at least begin date is mandatory ?).

Using the standard rendering, I notice that the labels are not displayed and both fields are set as mandatory, something to review in the `labels.xml` file.

[...]

PR #211, opened 2021-05-03
Validation error showing for temporal extend.

wangf1122:
Here is after the fix.

image

### DIFF
--- a/src/main/java/ca/gc/schema/iso19139hnap/util/XslUtilHnap.java
+++ b/src/main/java/ca/gc/schema/iso19139hnap/util/XslUtilHnap.java
@@ -123,8 +123,6 @@ public class XslUtilHnap {
 
         if (date.length() == 4) {
             try {
-                int yearValue = Integer.parseInt(date);
-
                 Date dateValue = DateUtils.parseIso8601DateTimeOrDate(date + "-01-01");
 
                 return 1;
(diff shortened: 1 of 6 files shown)


wangf1122:
Issue link:

https://github.com/metadata101/iso19139.ca.HNAP/issues/210

ianwallen (commented):
I don't believe we can remove the `templateModeOnly="true"`
When you do that the date selector is just a generic date selector.
We need the ability to be able to enter dates like the following.
image

wangf1122:
> I don't believe we can remove the `templateModeOnly="true"`
> When you do that the date selector is just a generic date selector.
> We need the ability to be able to enter dates like the following.
> image

@ianwallen 

The templateModeOnly has brought us another issue. The temporal element can never be removed. I think we should just get rid of that mode. Then add the date picker in this xsl. I had some luck to add it and it's doable. But will probably take long time. 

Can we do a short pull request to just give short relief (I will commit some code to this branch to force using datetime i.e. 2021-05-19T16:36:00)? Then we will investigate the date-picker option after?

wangf1122:
Come with final solution. The date-picker is moved into layout-custom-fields.xsl. It has the ability to swap among year, year-monty, and full date-time mode. Validation text works too.

image

josegar74:
The validation message works fine, but when I set a value, for example in the `Begin date` and click `Save metadata`, the following error is displayed:

error-save-temporal-extent

wangf1122:
@josegar74 

This is expected. We are passing string as is without node to <beginPosition> and <endPosition> as XMLFragment. So Geonetwork-Core should support String to parse. Here is the pull request for Geonetwork Core module. https://github.com/geonetwork/core-geonetwork/pull/5746

So they have dependency with each other.

jodygarnett:
Any update or discusion needed here?

wangf1122:
@jodygarnett 

This change has a dependency change in Geonetwork itself which is https://github.com/geonetwork/core-geonetwork/pull/5746. It is not approved yet. So we should not approve this change until Geonetwork community approves it.

josegar74:
@wangf1122 I see it's send this `_X249: 2021-07-26`. The field name starting with `_X` is used for xml fragments. Why is not send the value for the field like `_249: 2021-07-26` as other "simple" values?

It seems related to this: https://github.com/metadata101/iso19139.ca.HNAP/pull/211/files#diff-7a655927f8b146f5aeab45cbd707980efccd2f738f9a17c502592d19e3e9ae4fR85

Checking if configured like the following, it's not required the change in core GeoNetwork:

```
 data-element-ref="{concat('_', gn:element/@ref)}"
 ```

josegar74:
@wangf1122 I've tested previous change and looks ok, without requiring the change in https://github.com/geonetwork/core-geonetwork/pull/5746, can you double check?

wangf1122:
@josegar74 

I tested and confirmed the simple value mode  data-element-ref="{concat('_', gn:element/@ref)}"
  works. I will take https://github.com/geonetwork/core-geonetwork/pull/5746 off. Please just do merging for metadata101 part. Thanks

josegar74 (approved):

PR #216, opened 2021-05-13
Add Red Asterix Beside keyword

wangf1122:
Before:
image

After:
image


wangf1122:
@ianwallen 

I made similar change to the same xsl file as the orgName "red star" sign and check from label.xml if this field is mandatory or not.

PR #219, opened 2021-05-27
Add xlink:role attribute to gmd:online tag.

wangf1122:
The validation was giving such error:
image

The issue is xlink:role attribute was missing from gmd:online tag. So after the fix. The xlink:role will be added if no $language parameter passed into the xsl, we use root/language to determine what that language is. Here is the result transformed XML:

image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/process/onlinesrc-add.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/process/onlinesrc-add.xsl
@@ -178,9 +178,19 @@ Insert is made in first transferOptions found.
       <xsl:choose>
         <xsl:when test="starts-with($protocol, 'OGC:') and $name != ''">
           <gmd:onLine>
-            <xsl:if test="$language">
-              <xsl:attribute name="xlink:role" select="$language"/>
-            </xsl:if>
+            <xsl:choose>
+              <xsl:when test="$language">
+                <xsl:attribute name="xlink:role" select="$language"/>
+              </xsl:when>
+              <xsl:otherwise>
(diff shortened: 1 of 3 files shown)


akhamdo:
@jodygarnett 
Would it be possible to have a look as this is need for our next release?

wangf1122:
@jodygarnett 

Also to mention, the language is unique language and no requirement for the language at all. Just curious if we can remove that schematron rule as alternative solution too? Thanks.

josegar74:
Some comments about this attribute:

1) It was added to allow to select in the online resources dialog the resource language (this is a customisation done in the some versions of GeoNetwork), used to register the map services in RCS system, as when registering a map service required to be specified in both languages. This attribute was added as a convention, as there is no way to indicate a multilingual URL in `gmd:onLine`. 

2) With the current changes, it's added always with the metadata language, it's not a selection in the online resources dialog.

3) Not sure if RCS is used yet? If not, I think the best would be to get rid of this attribute and update the validation rules.

wangf1122:
Will take 3) approach to remove the validation rule. I committed the code. The validation rule is removed and I revert xlink:role back to the old version. Please take a look. @josegar74

josegar74:
Going to make a new PR, preserving the validation rules, but with this logic in the code to add online resources:

- if the description ends with fra add in xlink:role="urn:xml:lang:fra-CAN"
- If not (ie eng for fra-eng) add in xlink:role="urn:xml:lang:eng-CAN"

wangf1122:
This was replaced by https://github.com/metadata101/iso19139.ca.HNAP/pull/230.

PR #223, opened 2021-06-28
Convert fra to fre (ISO-639.2.B) for multilingual fields.

wangf1122:
The issue is Geonetwork uses language code "fre" which is ISO-639.2.B standard. NHAP uses "fra" as language code which is not fully supported by multilingual angular module in Geonetwork. There will be conversion needed within xsl.

Issue:
When the user select French NHAP profile to make metadata:
image

the system uses "eng" as default the multilingual field (for every single of them)
image

Problem is the language code is not consistent. It's mixed of ISO 6392T and ISO 6392B (fra and fre). This angular widget only support 6392B which is fre. 
image

So both main language and JSON needs to be consistent like the following.
image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
@@ -125,9 +125,15 @@
 
         <!-- the existing translation -->
         <xsl:for-each select="gmd:PT_FreeText/gmd:textGroup/gmd:LocalisedCharacterString">
+          <xsl:variable name="locale_ISO639_2B">
+            <xsl:choose>
+              <xsl:when test="@locale = '#fra'">#fre</xsl:when>
+              <xsl:otherwise><xsl:value-of select="@locale"/></xsl:otherwise>
+            </xsl:choose>
+          </xsl:variable>
           <!-- don't put in the default lang if it already has a value -->
-          <xsl:if test="not($hasDefaultValue) or substring-after(@locale, '#') != $metadataLanguage">
-            <value ref="{gn:element/@ref}" lang="{substring-after(@locale, '#')}">
+          <xsl:if test="not($hasDefaultValue) or substring-after($locale_ISO639_2B, '#') != $metadataLanguage">
+            <value ref="{gn:element/@ref}" lang="{substring-after($locale_ISO639_2B, '#')}">
               <xsl:value-of select="."/>
             </value>
           </xsl:if>
@@ -136,9 +142,16 @@
         <!-- and create field for none translated language -->
         <xsl:for-each select="$metadataOtherLanguages/lang">
           <xsl:variable name="currentLanguageId" select="@id"/>
+          <xsl:variable name="currentLanguageId_none_iso">
+            <xsl:choose>
+              <xsl:when test="$currentLanguageId = 'fre'">fra</xsl:when>
+              <xsl:otherwise><xsl:value-of select="$currentLanguageId"/></xsl:otherwise>
+            </xsl:choose>
+          </xsl:variable>
+
           <xsl:if test="count($theElement/
                 gmd:PT_FreeText/gmd:textGroup/
-                gmd:LocalisedCharacterString[@locale = concat('#',$currentLanguageId)]) = 0">
+                gmd:LocalisedCharacterString[@locale = concat('#',$currentLanguageId_none_iso)]) = 0">
             <!--don't put in default language if already there-->
               <xsl:if test="not($hasDefaultValue) or $currentLanguageId != $metadataLanguage ">
                  <value ref="lang_{@id}_{$theElement/gn:element/@ref}"
--- a/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
@@ -142,7 +142,7 @@
         <!-- and create field for none translated language -->
         <xsl:for-each select="$metadataOtherLanguages/lang">
           <xsl:variable name="currentLanguageId" select="@id"/>
-          <xsl:variable name="currentLanguageId_none_iso">
+          <xsl:variable name="currentLanguageId_iso639_2t">
             <xsl:choose>
               <xsl:when test="$currentLanguageId = 'fre'">fra</xsl:when>
               <xsl:otherwise><xsl:value-of select="$currentLanguageId"/></xsl:otherwise>
@@ -151,7 +151,7 @@
 
           <xsl:if test="count($theElement/
                 gmd:PT_FreeText/gmd:textGroup/
-                gmd:LocalisedCharacterString[@locale = concat('#',$currentLanguageId_none_iso)]) = 0">
+                gmd:LocalisedCharacterString[@locale = concat('#',$currentLanguageId_iso639_2t)]) = 0">
             <!--don't put in default language if already there-->
               <xsl:if test="not($hasDefaultValue) or $currentLanguageId != $metadataLanguage ">
                  <value ref="lang_{@id}_{$theElement/gn:element/@ref}"
--- a/src/main/plugin/iso19139.ca.HNAP/layout/layout.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/layout.xsl
@@ -231,7 +231,8 @@
           <xsl:for-each select="$metadataOtherLanguages/lang">
             <xsl:variable name="code" select="@code"/>
             <xsl:variable name="currentLanguageId" select="@id"/>
-            <xsl:variable name="currentLanguageId_none_ISO">
+
+            <xsl:variable name="currentLanguageId_iso639_2t">
               <xsl:choose>
                 <xsl:when test="$currentLanguageId = 'fre'">fra</xsl:when>
                 <xsl:otherwise><xsl:value-of select="$currentLanguageId"/></xsl:otherwise>
@@ -243,7 +244,7 @@
                           select="count($theElement/parent::node()/
                                         gmd:PT_FreeText/*/
                                         gmd:LocalisedCharacterString[
-                                          @locale = concat('#', $currentLanguageId_none_ISO)]) = 0"/>
+                                          @locale = concat('#', $currentLanguageId_iso639_2t)]) = 0"/>
 
 
             <xsl:choose>
--- a/src/main/plugin/iso19139.ca.HNAP/layout/layout.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/layout.xsl
@@ -214,8 +214,14 @@
 
           <!-- the existing translation -->
           <xsl:for-each select="gmd:PT_FreeText/gmd:textGroup/gmd:LocalisedCharacterString">
-            <xsl:if test="not($metadataLanguage = substring-after(@locale, '#'))">
-              <value ref="{gn:element/@ref}" lang="{substring-after(@locale, '#')}">
+            <xsl:variable name="locale_ISO639_2B">
+              <xsl:choose>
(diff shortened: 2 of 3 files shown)


src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl
```diff
@@ -102,7 +114,13 @@
...
                </xsl:choose>
             </xsl:variable>
-            <lang><xsl:value-of select="concat('&quot;', $lang_ISO639_2B, '&quot;:&quot;#', @id, '&quot;')"/></lang>
+            <xsl:variable name="id_ISO639_2B">
+              <xsl:choose>
+                <xsl:when test="@id = 'fra'">fre</xsl:when>
+                <xsl:otherwise><xsl:value-of select="@id"/></xsl:otherwise>
+              </xsl:choose>
+            </xsl:variable>
+            <lang><xsl:value-of select="concat('&quot;', $lang_ISO639_2B, '&quot;:&quot;#', $id_ISO639_2B, '&quot;')"/></lang>
```
ianwallen: I don't believe I agree with this change.
The `@id`  is just a reference to the PT_Locale.  It could be called `XYZ` for French local and referenced as `#XYZ` - if you need the real language code for  `#XYZ` then you have to get it from the codeValueList of the pt_locale.  
The `@id` should just be treated as a reference/variable name.

I have seen some logic in the code which treats the `@id` as the real language and it causes bugs. So if you see the  value from the `@id` treated as the real language somewhere, it is most likely a bug.

src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl
```diff
@@ -90,8 +96,14 @@
                   <xsl:otherwise><xsl:value-of select="$mainLanguage"/></xsl:otherwise>
                </xsl:choose>
             </xsl:variable>
+            <xsl:variable name="mainLanguageId_ISO639_2B">
```
ianwallen: Changing the value of an `@id` reference/variable name seems incorrect as this is only a reference to the PT_Locale.  It could be called `ABC`  for referencing the English PT_Locale.

src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl
fxprunayre: > The @id is just a reference to the PT_Locale. It could be called XYZ

True but note that if at some point you want to use multilingual subtemplates (which does not have a section to declare language ids) it will not work as it is now - that's why default is to build ids from the 2 first letters of the language code. We might be able to find more elegant solution for this BTW ... It mainly comes from geocat.ch in which user can only use subtemplates for contacts, extents and formats.

wangf1122: I have fixed the code to use code list instead of id here. I also added some more code fix to layout.xsl which is responsible for this element in English mode
<gmd:PT_FreeText>
       <gmd:textGroup>
           <gmd:LocalisedCharacterString locale="#fra">200 rue Kent</gmd:LocalisedCharacterString>
        </gmd:textGroup>
 </gmd:PT_FreeText>

src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl
```diff
@@ -47,7 +47,15 @@
             <xsl:value-of select="$metadata/gmd:language/gmd:LanguageCode/@codeListValue"/>
           </xsl:when>
           <xsl:when test="contains($metadata/gmd:language/gco:CharacterString,';')">
-            <xsl:value-of  select="normalize-space(substring-before($metadata/gmd:language/gco:CharacterString,';'))"/>
+            <xsl:variable name="metadataMainLanguage" select="normalize-space(substring-before($metadata/gmd:language/gco:CharacterString,';'))"/>
```
ianwallen: There is a function for getting the main language and it does more than just taking the language before the ; - it would be better if you used that function if possible.

src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl
```diff
@@ -112,7 +137,51 @@
...
+      <xsl:otherwise>
+        <xsl:variable name="mainLanguage">
+          <xsl:call-template name="get-iso19139-language"/>
+        </xsl:variable>
+        <xsl:for-each select="$metadata/gmd:locale/gmd:PT_Locale">
+          <xsl:variable name="langCode"
+                        select="gmd:languageCode/gmd:LanguageCode/@codeListValue"/>
+          <xsl:variable name="id_ISO639_2B">
+            <xsl:choose>
+              <xsl:when test="@id='fra'">
```
ianwallen: Should not be comparing to the @id as it is a variable name?

src/main/plugin/iso19139.ca.HNAP/layout/layout.xsl
```diff
@@ -225,13 +231,19 @@
           <xsl:for-each select="$metadataOtherLanguages/lang">
             <xsl:variable name="code" select="@code"/>
             <xsl:variable name="currentLanguageId" select="@id"/>
+            <xsl:variable name="currentLanguageId_none_ISO">
```
ianwallen: Odd to use none_ISO for a variable name.  Why not rename the other variable to ISO?

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
```diff
@@ -136,9 +142,16 @@
...
+            <xsl:choose>
+              <xsl:when test="$currentLanguageId = 'fre'">fra</xsl:when>
+              <xsl:otherwise><xsl:value-of select="$currentLanguageId"/></xsl:otherwise>
+            </xsl:choose>
+          </xsl:variable>
+
           <xsl:if test="count($theElement/
                 gmd:PT_FreeText/gmd:textGroup/
-                gmd:LocalisedCharacterString[@locale = concat('#',$currentLanguageId)]) = 0">
+                gmd:LocalisedCharacterString[@locale = concat('#',$currentLanguageId_none_iso)]) = 0">
```
ianwallen: Odd to use none_ISO for a variable name. Why not rename the other variable to ISO?

wangf1122: I thought about it but the other variable will the locale attribute and it's on the xpath. So would this be better?

<variable name = "local_ISO" select ="$theElement/gmd:PT_FreeText/gmd:textGroup/gmd:LocalisedCharacterString/@locale">?

Then use this to compare?

Edited: I just realized why I used noneISO instead. The function is inside a document scan with counting with xpath:

<xsl:variable name="ptFreeElementDoesNotExist"
                          select="count($theElement/parent::node()/
                                        gmd:PT_FreeText/*/
                                        gmd:LocalisedCharacterString[
                                          @locale = concat('#', $currentLanguageId_none_ISO)]) = 0"/>

There is no way I could do locale converting here. I am not objecting to use a variable called ISO but how it can be part of the count function? Please advise.

src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl
wangf1122: I was following what's been done and attempt to make minimum change. Which function are you referring to by the way?

src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl
wangf1122: I should not have this variable at all. The id should be passed from ISO converted codeList which is $langCode_ISO639_2B.

src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl
ianwallen: I'm hoping one of these would work.

Here are 2 from HNAP

https://github.com/metadata101/iso19139.ca.HNAP/blob/564a276ecd1d6d53cef7aad640ce2ce84bfa7056/src/main/plugin/iso19139.ca.HNAP/convert/functions.xsl#L39-L55

https://github.com/metadata101/iso19139.ca.HNAP/blob/564a276ecd1d6d53cef7aad640ce2ce84bfa7056/src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl#L38-L58

Here and similar from iso19139

https://github.com/geonetwork/core-geonetwork/blob/ce3d7d5f88aeb58404b695c8c4ed8142a78a1e4c/schemas/iso19139/src/main/plugin/iso19139/convert/functions.xsl#L149-L165

https://github.com/geonetwork/core-geonetwork/blob/ce3d7d5f88aeb58404b695c8c4ed8142a78a1e4c/schemas/iso19139/src/main/plugin/iso19139/layout/utility-tpl-multilingual.xsl#L35-L55

src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl
ianwallen: So why are you converting an id from fra to fre.  If the id in the xml is called `fra` then it will be referenced as `#fra` so changine the variable to fre seems incorrect.  If you want fre then it seems like it is using langCode_ISO639_2B. I'm not sure I follow the logic of changing variable names in the code?

wangf1122: This whole function was copied from ISO19139 profile and I attempt to convert both of its id and value to fre:#fre. The key:value will be passed to other module as I understand. I don't the full context of that id, but if the id is passed as fra, then we will have issue in English mode when the other language (French) cannot be detected.

src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl
wangf1122: Confirmed, it works. I will commit the code for it.

src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl
```diff
@@ -47,7 +48,19 @@
             <xsl:value-of select="$metadata/gmd:language/gmd:LanguageCode/@codeListValue"/>
           </xsl:when>
           <xsl:when test="contains($metadata/gmd:language/gco:CharacterString,';')">
-            <xsl:value-of  select="normalize-space(substring-before($metadata/gmd:language/gco:CharacterString,';'))"/>
+            <xsl:variable name="metadataMainLanguage">
+              <xsl:call-template name="langId_from_gmdlanguage19139">
+                <xsl:with-param name="gmdlanguage" select="/root/*/gmd:language"/>
+              </xsl:call-template>
+            </xsl:variable>
+            <xsl:choose>
```
wangf1122, line 56: This choose clause can probably call a function because it's been used multiple times within this code change.

src/main/plugin/iso19139.ca.HNAP/layout/layout.xsl
wangf1122: I have rename the variable to currentLanguageId_iso639_2t to indicate it is iso639-2t format.

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
wangf1122: I have rename the variable to currentLanguageId_iso639_2t to indicate it is iso639-2t format.

ianwallen (approved):

PR #227, opened 2021-07-09
Validation - check the format of the reference system code value

josegar74:


jodygarnett:
I have tested locally, thanks for the improvement:

>  **Metadata Information**
>  Reference System Code should have the format EPSG:NUMBER / SR-ORG:NUMBER or have the value Proj4. Example: EPSG:4326

PR #228, opened 2021-07-09
Change gmd:spatialRepresentationType to optional and remove validation rules

wangf1122:
MD_DataIdentification/gmd:spatialRepresentationType is current marked as mandatory. But accordingly to the foot note of related document, it meant to be optional
image

I have commit the solution to get this fields converted to optional and removing validation rule of missing. Please verify if this is acceptable.


jodygarnett:
Tested locally, thank you for the improvement.

jodygarnett:
Backported to 3.10.7 (sadly we cannot have the same PR against two milestones)

PR #234, opened 2021-08-30
Use default version from its parent in pom.xml

wangf1122:
The version is always same as its parent. It's better to remove it and use the default from its parent. Should keep same dependency management mechanism as other major metadata profile (i,.e ISO 19139).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -8,7 +8,7 @@
     <version>3.12-SNAPSHOT</version>
   </parent>
   <modelVersion>4.0.0</modelVersion>
-  <version>3.12-SNAPSHOT</version>
+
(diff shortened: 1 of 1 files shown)


wangf1122:
Hi @jodygarnett 

Do you have any thing to add here if I remove this version or you prefer to let it stay?

ianwallen (approved):

ianwallen:
@jodygarnett you had added this in past revision but don't believe it is required.  If you do no object, I will merge this change.

jodygarnett:
Go ahead and remove; I agree that the version comes from the parent.

jodygarnett (approved):

PR #235, opened 2021-08-31
Add missing alias attribute to codelists

josegar74:
Some UI angularjs directives like the directory entry selector to add contacts in the metadata editor, use the `alias` attribute from the codelist file to retrieve the values, instead of using the element name, probably to make them more generic to work with other schemas like `iso19115-3` that use different element names.

The `alias` attribute was missing in HNAP, causing that the values default to the ones from `iso19139`, but the codes in HNAP use a different format, causing that the elements ended with the codes from `iso19139` instead of the HNAP ones.

Test case:

1) Import both iso19139 and HNAP templates

2) Create a contact directory entry

3) Create an HNAP metadata and select the contact from the directory list, selecting a role

Without this fix, the role ends with the iso19139 code value, for example `owner`. WIth the fix, for the owner role it's assigned the code from HNAP `RI_410`.


jodygarnett (approved):
thanks this worked great

PR #240, opened 2021-09-30
3.12.x.default.language.multilingual.fields

wangf1122:


ianwallen:
@josegar74 

It seems like a simpler fix would be to modify the following.

https://github.com/geonetwork/core-geonetwork/blob/9504f940bcea08abe6aff36eecda907ae99c57cf/web-ui/src/main/resources/catalog/components/edit/multilingualfield/MultilingualFieldDirective.js#L75

To something like the following 
```

          // Get languages from attributes (could be grab from the
          // form field ? FIXME)

          if (scope.mainLanguage == 'fra')
            scope.mainLanguage='fre'

          scope.languages = angular.fromJson(attrs.gnMultilingualField);
          var mainLanguage = scope.mainLanguage;

```

But converting fra to fre in this locations does not seem to be the best way to fix the issue.

I cannot identify where "scope.mainLanguage" is being set. If we can locate this then we could apply the new iso639_2T_to_iso639_2B function (https://github.com/geonetwork/core-geonetwork/pull/5982) . – I believe this would be a better fix.

Any suggestions as to where  "scope.mainLanguage" is being set?

josegar74:
Relevant code:

- Form builder, add `metadataLanguage` to multilingual elements: https://github.com/geonetwork/core-geonetwork/blob/e648c60da714d40527dcbb743fcfa376a789b15a/web/src/main/webapp/xslt/ui-metadata/form-builder.xsl#L214

- Code to calculate `metadataLanguage` variable: https://github.com/geonetwork/core-geonetwork/blob/e648c60da714d40527dcbb743fcfa376a789b15a/web/src/main/webapp/xslt/common/base-variables-metadata.xsl#L58-L60

- HNAP `metadataLanguage` template: https://github.com/metadata101/iso19139.ca.HNAP/blob/b37d4f95091af69974c0036d05f5100a28a6db92/src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl#L37-L58

wangf1122:
> Relevant code:
> 
> * Form builder, add `metadataLanguage` to multilingual elements: https://github.com/geonetwork/core-geonetwork/blob/e648c60da714d40527dcbb743fcfa376a789b15a/web/src/main/webapp/xslt/ui-metadata/form-builder.xsl#L214
> * Code to calculate `metadataLanguage` variable: https://github.com/geonetwork/core-geonetwork/blob/e648c60da714d40527dcbb743fcfa376a789b15a/web/src/main/webapp/xslt/common/base-variables-metadata.xsl#L58-L60
> * HNAP `metadataLanguage` template: https://github.com/metadata101/iso19139.ca.HNAP/blob/b37d4f95091af69974c0036d05f5100a28a6db92/src/main/plugin/iso19139.ca.HNAP/layout/utility-tpl-multilingual.xsl#L37-L58

I confirmed we can make a simple fix to Geonetwork form-builder.xsl with language conversion. This pull request should be abandoned. 

Here is the replacement PR: https://github.com/geonetwork/core-geonetwork/pull/6024

wangf1122:
PR no longer needed as replaced by https://github.com/geonetwork/core-geonetwork/pull/6024

PR #242, opened 2021-10-06
Use new getDefaultLangCode to get the default language

ianwallen:
Once the following PR has been back ported to 3.12.x

https://github.com/geonetwork/core-geonetwork/pull/5981

Then the these changes will be required to get the default language


juanluisrp:
geonetwork/core-geonetwork#5981 has been backported to `3.10.x` and `3.12.x`

PR #250, opened 2021-12-23
Avoid indexing srv:serviceType/gco:LocalName with empty values

josegar74:
Otherwise if the empty value is in a metadata template, the metadata new record page shows an entry `service-` for the template types:

template-type


jodygarnett (approved):
looks good to me

PR #251, opened 2022-01-03
Remove role validation from multiligual rules

josegar74:
The validation is defined in the common rules and was reporting an empty error message for the multilingual rules:

validation


jodygarnett:
Is the solution to remove the role validation check, or provide it with an error message to report?

ianwallen:
I agree with @josegar74 fix as roles are not multilingual

And the fix would be more consistent with other role checks which only seem to check common.

image

ianwallen (approved):

PR #254, opened 2022-01-18
Fix display of multilingual French values in the full view formatter

josegar74:
WIthout this change the full view formatter displays in the French UI, English values for multilingual elements, due to the wrong language code used.

language

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/formatter/xsl-view/view.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/formatter/xsl-view/view.xsl
@@ -53,6 +53,8 @@
   <xsl:variable name="editorConfig"
                 select="document('../../layout/config-editor.xml')"/>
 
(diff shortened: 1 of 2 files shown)


src/main/plugin/iso19139.ca.HNAP/formatter/xsl-view/view.xsl
```diff
@@ -52,8 +53,10 @@
   <xsl:variable name="editorConfig"
                 select="document('../../layout/config-editor.xml')"/>
 
-  
-  <!-- Override codelist template, to don't use the text value, 
+  <xsl:variable name="langId" select="gn-fn-iso19139:getLangIdHNAP($metadata, $language)"/>
```
ianwallen, line 56: Would it not be better to use something like the following instead of creating a new gn-fn-iso19139:getLangIdHNAP function?

`gn-fn-iso19139:getLangId($metadata, xslUtils:iso639_2B_to_iso639_2T($language))`

PR #255, opened 2022-01-21
Fix country validation rules to verify the value in the thesaurus using the element language

josegar74:
Previously the values were check in both languages in the thesaurus, causing that this kind of entries (English element with the French value) passed the validation:

```
<gmd:country xsi:type="gmd:PT_FreeText_PropertyType">
    <gco:CharacterString>Canada (le)</gco:CharacterString>
    <gmd:PT_FreeText>
        <gmd:textGroup>
            <gmd:LocalisedCharacterString xmlns="" locale="#fra">Canada (le)</gmd:LocalisedCharacterString>
        </gmd:textGroup>
    </gmd:PT_FreeText>
</gmd:country>
```


ianwallen (approved):
LGTM

PR #256, opened 2022-01-31
Change label "systems" to "system" in singular word.

wangf1122:
The issue was the label for reference system. There is only one single system input but the "systems" word is multiplier.

image

It's better to change to singular word.

image


ianwallen:
I'm not sure I agree with this change.

The same text is used in the advanced view where we can add multiple reference system**s** as show here

image

However I will admit that the wording on these reference tags are a little misleading

"Ref. System Reference System" is odd but I think it is caused by the way the XML is layed out.

wangf1122:
Close this PR as it is not functional requirement.

PR #258, opened 2022-02-01
Include the resource url in the resource validation error messages

josegar74:
Include the online resource url in the validation messages to facilitate the identification of the resource with the invalid information:

resources-validation-messages

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
@@ -138,6 +138,18 @@
     </xsl:for-each>
   </xsl:function>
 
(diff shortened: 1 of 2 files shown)


jodygarnett:
That looks very useful @josegar74 !

ianwallen (approved):
@josegar74 
Changes look good 

I was going to recommend that the resource name be used instead of the link but then noticed that the resource name is not mandatory.

I also had a issue recently where I got this error but I could not identify the value that was invalid (I think it was an older record and I did not have time to debug so I just created a new one). It had looked valid but it was failing. It is possible that I was looking at the wrong item so adding the link will definitely help.  At some point, we may also want to look at including the value that was invalid in the error just so we can see the invalid value.

Update: looks like the issue I was having was related to the following #259

PR #261, opened 2022-02-14
Add priority to extract-relations template

josegar74:
Adding priority to the `get-iso19139-localized-string`.

Unless misunderstanding the xsl spec, using `xsl:import` allows to override templates, but we have identified in some deployments it's not the case. Adding the priority to make sure for HNAP metadata is used the correct template.


juanluisrp (approved):
Tested. LGTM.

PR #263, opened 2022-03-10
validate for online resource protocol

wangf1122:
The issue is this field is mandatory and can still be saved as empty from the user input.

image

As result, the protocol will be filled with some default value "WWW:DOWNLOAD-1.0-http--download" which is not correct. There is correct value which is HTTP as quick compare.

image

The validation rule will catch this default incorrect value and output error message as:
image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
@@ -41,7 +41,8 @@
   <DistributionFormatVersion>Value is required for distribution format - File version</DistributionFormatVersion>
   <DistributionFormatInvalid>Value for distribution format is not valid</DistributionFormatInvalid>
   <OnlineResourceUrl>Value is required for Online Resource URL</OnlineResourceUrl>
-  <OnlineResourceProtocol>Protocol (HTTP, HTTPS, FTP, ESRI REST: Map Service, OGC:WMS, OGC:WFS) is required for Online Resource: </OnlineResourceProtocol>
+  <OnlineResourceName>Protocol of Online Resource "</OnlineResourceName>
+  <OnlineResourceProtocol>" is not valid. Valid values are: </OnlineResourceProtocol>
 
   <MapResourcesREST>REST Map Resources should be available in english and french</MapResourcesREST>
   <MapResourcesRESTNumber>The number of ESRI REST Map Resources should be at most 2 (one for english and one for french)</MapResourcesRESTNumber>
--- a/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
@@ -41,6 +41,7 @@
   <DistributionFormatVersion>Value is required for distribution format - File version</DistributionFormatVersion>
   <DistributionFormatInvalid>Value for distribution format is not valid</DistributionFormatInvalid>
   <OnlineResourceUrl>Value is required for Online Resource URL</OnlineResourceUrl>
+  <OnlineResourceProtocol>Protocol is required for Online Resource: </OnlineResourceProtocol>
 
   <MapResourcesREST>REST Map Resources should be available in english and french</MapResourcesREST>
   <MapResourcesRESTNumber>The number of ESRI REST Map Resources should be at most 2 (one for english and one for french)</MapResourcesRESTNumber>
--- a/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
@@ -41,8 +41,7 @@
   <DistributionFormatVersion>Value is required for distribution format - File version</DistributionFormatVersion>
   <DistributionFormatInvalid>Value for distribution format is not valid</DistributionFormatInvalid>
   <OnlineResourceUrl>Value is required for Online Resource URL</OnlineResourceUrl>
-  <OnlineResourceName>Protocol of Online Resource "</OnlineResourceName>
-  <OnlineResourceProtocol>" is not valid. Valid values are: </OnlineResourceProtocol>
+  <OnlineResourceProtocol>Protocol of Online Resource is not valid. Valid values are: </OnlineResourceProtocol>
 
   <MapResourcesREST>REST Map Resources should be available in english and french</MapResourcesREST>
   <MapResourcesRESTNumber>The number of ESRI REST Map Resources should be at most 2 (one for english and one for french)</MapResourcesRESTNumber>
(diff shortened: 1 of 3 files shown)


ianwallen (approved):

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -403,6 +403,14 @@
 
     </sch:rule>
 
+    <sch:rule context="//gmd:distributionInfo/gmd:MD_Distribution/gmd:transferOptions/gmd:MD_DigitalTransferOptions/gmd:onLine/gmd:CI_OnlineResource">
+        <sch:let name="protocolMissing" value="not(string(gmd:protocol/gco:CharacterString))" />
+    		<sch:let name="isValidProtocol" value="(string(gmd:protocol/gco:CharacterString) != 'WWW:DOWNLOAD-1.0-http--download')"/>
```
josegar74: Maybe to validate these values if are the only ones allowed? 

https://github.com/metadata101/iso19139.ca.HNAP/blob/5da459ca3a777956d3b3cf6bf7d05186711ac500/src/main/plugin/iso19139.ca.HNAP/loc/eng/labels.xml#L1751-L1756

I'm not sure if from schematron file can access the `labels.xml` file, but otherwise maybe as are few values, hardcode them in the rule?

wangf1122: I cannot find any way to do it from labels.xml. So have to hardcode it. I also updated the error message to indicate

image


But in reality, this is from a selection box and there is no way user can input some "bad" selection.

wangf1122:
@josegar74 @ianwallen 

This seems to be a pending PR which we skipped.

ianwallen:
@wangf1122 
I can see in the compiled .xslt (from .sch file) that there is a line similar to the following to get the schematron locale data.

```
  <xsl:variable xmlns:svrl="http://purl.oclc.org/dsdl/svrl" name="loc"
                 select="document(concat('../loc/', $lang, '/', $rule, '.xml'))"/>
```

So I would expect that you should be able to do the following to get the labels.xml locales file

`<xsl:variable name="locLabel" select="document(concat('../loc/', $lang, '/labels.xml'))"/>`

In schematron rules, you cannot use `<xsl:variable` but you can use `<sch:let`  and you can also use `<xsl:function`.  So you can try with `<sch:let`. If you can't get `<sch:let` to work then you can try to create a `<xsl:function` to be used in the `<sch:let`.  Look at the other schematrons for sample using `<xsl:function`

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -403,6 +403,21 @@
 
     </sch:rule>
 
+    <sch:rule context="//gmd:distributionInfo/gmd:MD_Distribution/gmd:transferOptions/gmd:MD_DigitalTransferOptions/gmd:onLine/gmd:CI_OnlineResource">
+        <sch:let name="protocolMissing" value="not(string(gmd:protocol/gco:CharacterString))" />
+    		<sch:let name="isNotPlaceholderProtocol" value="(string(gmd:protocol/gco:CharacterString) != 'WWW:DOWNLOAD-1.0-http--download')"/>
```
josegar74: I think this check can be removed, no? Below are checked the the valid values.

wangf1122: I totally refined the validation. I used @ianwallen suggested approach (retrieve protocol values from label.xml) and got good result. Here is what it looks likes from my debug console.

image

src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
```diff
@@ -41,6 +41,7 @@
   <DistributionFormatVersion>Value is required for distribution format - File version</DistributionFormatVersion>
   <DistributionFormatInvalid>Value for distribution format is not valid</DistributionFormatInvalid>
   <OnlineResourceUrl>Value is required for Online Resource URL</OnlineResourceUrl>
+  <OnlineResourceProtocol>Protocol (HTTP, HTTPS, FTP, ESRI REST: Map Service, OGC:WMS, OGC:WFS) is required for Online Resource: </OnlineResourceProtocol>
```
ianwallen: @wangf1122 
Can you change this to be similar to SecurityLevel which has something like the following.

`<SecurityLevel>Security User Note is not valid. Valid values are:</SecurityLevel>`

Then it list the values when generating the error.  This would eliminated hardcoding the list of protocols.

wangf1122: Code updated. The result is:

image

src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
```diff
@@ -41,6 +41,8 @@
   <DistributionFormatVersion>Value is required for distribution format - File version</DistributionFormatVersion>
   <DistributionFormatInvalid>Value for distribution format is not valid</DistributionFormatInvalid>
   <OnlineResourceUrl>Value is required for Online Resource URL</OnlineResourceUrl>
+  <OnlineResourceName>Protocol of Online Resource "</OnlineResourceName>
+  <OnlineResourceProtocol>" is not valid. Valid values are: </OnlineResourceProtocol>
```
ianwallen: I'm not a fan of splitting the message into to parts. It makes the second part confusing.
I think it would be better to use a placeholder

i.e. Protocol for Online Resource "{invalidProtocol}" is not valid. Valid values are: 

The placeholder can be anything as long as it is not text that could appear in the message.

And then search and replace the placeholder.  This makes it easier to upgrade the text in the future and it also makes it easier to understand/read the full message.

josegar74 (approved):

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -403,6 +429,22 @@
...
+        <sch:let name="protocolList" value="$locLabel/labels/element[@name='gmd:protocol']/helper/option"/>
+        <sch:let name="protocol" value="gmd:protocol/gco:CharacterString/text()"/>
+        <sch:let name="isValidProtocol" value="$protocol = $protocolList"/>
+
+    		<sch:let name="resourceName" value="gmd:name/gco:CharacterString/text()" />
+
+    		<sch:let name="protocolListString" value="geonet:protocolListString($protocolList)"/>
+
+    		<sch:let name="protocolListMsg" value="geonet:appendLocaleMessage($loc/strings/OnlineResourceProtocol, $protocolListString)"/>
+    		<sch:let name="locMsg" value="geonet:appendLocaleMessage($resourceNameMsg, $protocolListMsg)"/>
```
josegar74, line 443: @wangf1122 where is defined `$resourceNameMsg`? Seem it's missing in the PR, we just notice doing further testing. 

Should be instead `resourceName`?

PR #264, opened 2022-04-08
Configure the online resources dialog to display the layer selector for WMS layers and add the layer names in the service url. 

josegar74:
Related to #6195

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/config/associated-panel/default.json
+++ b/src/main/plugin/iso19139.ca.HNAP/config/associated-panel/default.json
@@ -1,7 +1,7 @@
 {
   "config": {
     "display": "radio",
-    "loadMapCapabilities": "false",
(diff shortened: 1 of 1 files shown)


wangf1122:
I tested from my local and got this error.

image


It's possibly caused by this line of code:
https://github.com/geonetwork/core-geonetwork/blob/af5c60071a9013e23e18217b1551b28e8b8896a0/web-ui/src/main/resources/catalog/components/edit/onlinesrc/OnlineSrcDirective.js#L706

The urlLayerParam is not part of JSON configuration.

I think it had some dependency with this commit:
https://github.com/geonetwork/core-geonetwork/commit/ec40def2501f7bc79fd3579aa9ef57d89a722bec

After I cherry pick this commit, it works fine 
image

ianwallen (approved):

PR #269, opened 2022-05-13
dataQuality/scope not mandatory

wangf1122:
When adding dataset link

image

the dataquality node was added like this:
image

The schema validation failed
image

The xsl which is in ISO19139 has no such gmd:scope related code:
https://github.com/geonetwork/core-geonetwork/blob/29f9f5e3b9bc390d82e23d75efe29d9a757c8e72/schemas/iso19139/src/main/plugin/iso19139/process/source-add.xsl#L73-L83

So I would like to make this gmd:scope not mandatory in its schema.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schema/gmd/dataQuality.xsd
+++ b/src/main/plugin/iso19139.ca.HNAP/schema/gmd/dataQuality.xsd
@@ -502,7 +502,7 @@
 		<xs:complexContent>
 			<xs:extension base="gco:AbstractObject_Type">
 				<xs:sequence>
-					<xs:element name="scope" type="gmd:DQ_Scope_PropertyType"/>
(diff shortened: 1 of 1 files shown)


ianwallen:
@wangf1122 - I don't believe we want to be modifying the HNAP schema specification as we are not the authoritive source for this data.

I think the fix should be to modifying the following
https://github.com/metadata101/iso19139.ca.HNAP/blob/3.12.x/src/main/plugin/iso19139.ca.HNAP/process/source-add.xsl
so that it does not use the iso19139.ca's version and instead adds it own custom version with the gmd:scope

@josegar74 - Any suggestion?

josegar74:
I think it should be fixed the xsl, to add the `<gmd:scope />` element instead of changing the `xsd`. If the element content is not relevant, can be added with empty content, as `<gmd:scope />`, what is valid.

wangf1122:
This PR is no longer needed and replaced by https://github.com/metadata101/iso19139.ca.HNAP/pull/331

PR #272, opened 2022-06-15
Metadata editor - data identification dates - support only date format

josegar74:
Requires https://github.com/geonetwork/core-geonetwork/pull/6394

HNAP requires date format only for the data identification dates, with this change the fields in the editor display allow to select only dates in the metadata editor and does not allow to change to year/month or year modes.

datepicker-dateonly

A validation rule is added also to validate the full date format: `YYYY-MM-DD`:

validation-rule

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
@@ -18,6 +18,7 @@
   <PublicationDate>Value is required for Metadata publication date</PublicationDate>
   <CreationDate>Value is required for Metadata creation date</CreationDate>
   <MissingDate>Value is required for Date</MissingDate>
+  <InvalidDate>Date format is not valid. Should be: YYYY-MM-DD</InvalidDate>
   <InvalidDateTypeCode>Date type is not valid</InvalidDateTypeCode>
   <BeginDate>Value is required for Begin Date</BeginDate>
   <BeginPositionFormat>Begin position format is not valid value</BeginPositionFormat>
--- a/src/main/plugin/iso19139.ca.HNAP/process/fix-date-format.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/process/fix-date-format.xsl
@@ -1,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xsl:stylesheet   xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="2.0"
+                  xmlns:gco="http://www.isotc211.org/2005/gco"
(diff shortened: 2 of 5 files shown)


src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
```diff
@@ -18,6 +18,7 @@
   <PublicationDate>Value is required for Metadata publication date</PublicationDate>
   <CreationDate>Value is required for Metadata creation date</CreationDate>
   <MissingDate>Value is required for Date</MissingDate>
+  <InvalidDate>Date format is not valid. Should be: YYYY-MM-DD</InvalidDate>
```
ianwallen, line 21: As discussed (and shown in your screenshot) the date format can be in different formats depending on locale so maybe this message should be changed to be more generic.

src/main/plugin/iso19139.ca.HNAP/process/fix-date-format.xsl
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<xsl:stylesheet   xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="2.0"
+                  xmlns:gco="http://www.isotc211.org/2005/gco"
+                  xmlns:gmd="http://www.isotc211.org/2005/gmd"
+                  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+                  xmlns:geonet="http://www.fao.org/geonetwork"
+                  exclude-result-prefixes="#all">
+
+  <xsl:template match="gmd:MD_DataIdentification/gmd:citation/gmd:CI_Citation/gmd:date/gmd:CI_Date/gmd:date[gco:Date and string-length(gco:Date) != 10]/gco:Date">
```
ianwallen, line 9: Question?

Is this required for the following as well?

`gmd:identificationInfo/*/gmd:extent/gmd:EX_Extent/gmd:temporalElement/gmd:EX_TemporalExtent/gmd:extent/gml:TimePeriod/gml:beginPosition`

We currently have issues when publishing this date to Open Data.  It does not support single date as a year so we have to add the 01-01 as well.  It is possible that this is an Open Data issues and not necessarily an HNAP issue so I'm just asking the question.

src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
jodygarnett: Suggest:

> The date is entered using your current locale, it will store the date in YYYY-MM-DD format

jodygarnett: Alternate:

> The date entered is not in the correct format for your locale

PR #273, opened 2022-06-27
Fix typo

wangf1122:
There is a typo in services-remove.xsl filename.

image

Therefore the process is not found by Geonetwork. Also there the base xsl is named services.xml https://github.com/geonetwork/core-geonetwork/blob/main/schemas/iso19139/src/main/plugin/iso19139/process/services-remove.xsl

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/process/services-remove.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/process/services-remove.xsl
@@ -23,5 +23,5 @@
   -->
 
 <xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="2.0">
-  <xsl:import href="../../iso19139/process/service-remove.xsl"/>
+  <xsl:import href="../../iso19139/process/services-remove.xsl"/>
 </xsl:stylesheet>


ianwallen:
Odd that HNAP uses "service-remove.xsl" (no "s")

https://github.com/metadata101/iso19139.ca.HNAP/blob/3.12.x/src/main/plugin/iso19139.ca.HNAP/process/service-remove.xsl

But ISO 19139 uses "services-remove.xsl" (with "s")

https://github.com/geonetwork/core-geonetwork/blob/main/schemas/iso19139/src/main/plugin/iso19139/process/services-remove.xsl

So not sure if we should also rename the HNAP version to match iso19139 or if it should be changed in ISO 19139?

It seems like the plural is used from some of the removal xslt i.e. [datasets-remove.xsl](https://github.com/metadata101/iso19139.ca.HNAP/blob/3.12.x/src/main/plugin/iso19139.ca.HNAP/process/datasets-remove.xsl)?

@josegar74 do you have any suggestions on what should be done for consistency?

josegar74 (approved):

ianwallen:
@wangf1122, could you also rename the following to include the "S" so that it matches the iso19139 version https://github.com/metadata101/iso19139.ca.HNAP/blob/3.12.x/src/main/plugin/iso19139.ca.HNAP/process/service-remove.xsl

ianwallen:
@wangf1122 sorry - I see that you already renamed the file.  I has missed it when looking at the changes.  

I will approve.

ianwallen (approved):

PR #274, opened 2022-06-29
Use HTTP protocol for DOI links. DOI protocol is not supported in HNAP

josegar74:


src/main/plugin/iso19139.ca.HNAP/process/doi-add.xsl
```diff
@@ -20,6 +20,9 @@
    </xsl:choose>
   </xsl:variable>
 
+  <xsl:variable name="doiProtocol"
+                select="'HTTP'"/>
```
ianwallen: Is there a reason that the protocol is hard coded to HTTP vs HTTPS?   Most apps are now using HTTPS.

Is it possible to extract the protocol from the doi url?

josegar74: @ianwallen updated the code, please check.

wangf1122 (approved):

ianwallen (approved):

PR #276, opened 2022-07-04
Add validation for srv:SV_ServiceIdentification

wangf1122:
This legacy service type metadata do existed when we imported from other system.

image

The <gmd:MD_DataIdentification> is only for dataset type not service. Service metadata will have srv:SV_ServiceIdentification element name instead. This validation rule just to ensure srv:SV_ServiceIdentification must be presented with gmd:MD_Metadata/gmd:hierarchyLevel/gmd:MD_ScopeCode='service'


josegar74 (approved):
Tested, works fine

PR #282, opened 2022-08-19
Support for a process to update the metadata content when creating a duplicate

josegar74:
This change requires https://github.com/geonetwork/core-geonetwork/pull/6507 to be back ported to `3.12.x` branch.

Currently, it is used to remove DOI information when duplicating or creating a child from a metadata, but can be extended with additional use cases if needed.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/duplicate-metadata.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/duplicate-metadata.xsl
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright (C) 2001-2016 Food and Agriculture Organization of the
+  ~ United Nations (FAO-UN), United Nations World Food Programme (WFP)
+  ~ and United Nations Environment Programme (UNEP)
+  ~
+  ~ This program is free software; you can redistribute it and/or modify
+  ~ it under the terms of the GNU General Public License as published by
+  ~ the Free Software Foundation; either version 2 of the License, or (at
+  ~ your option) any later version.
+  ~
+  ~ This program is distributed in the hope that it will be useful, but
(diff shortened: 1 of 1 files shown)


src/main/plugin/iso19139.ca.HNAP/duplicate-metadata.xsl
```diff
@@ -0,0 +1,57 @@
...
+
+
+  <!-- Remove DOI identifiers -->
+  <xsl:template match="gmd:identifier[
+                                contains(*/gmd:code/*/text(), 'datacite.org/doi/')
+                                or contains(*/gmd:code/*/text(), 'doi.org')
+                                or contains(*/gmd:code/*/@xlink:href, 'doi.org')]" />
+
+  <!-- Remove DOI links -->
+  <xsl:template match="gmd:onLine[*/gmd:name/gco:CharacterString = $doiName]" />
```
ianwallen, line 54: Could you add a comment indicating why this was not simply an import like the following and why we are not matching based on cit:protocol like in iso19139.
 
```
<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform" version="2.0">
  <xsl:import href="../iso19139/duplicate-metadata.xsl"/>
</xsl:stylesheet>
```

ianwallen (approved):
Approved on the condition that https://github.com/geonetwork/core-geonetwork/pull/6507 is approved as well and that it has been backported to 3.12.x.

I also made some suggestions for some comments.

src/main/plugin/iso19139.ca.HNAP/duplicate-metadata.xsl
josegar74: @ianwallen added a comment to explain the custom HNAP change.

ianwallen (approved):

PR #283, opened 2022-08-26
Add schematron validation rules for datacite

josegar74:


ianwallen:
@josegar74 could provide some context on this requirement?

I would like to understand why this is added as a separate schematron instead of being added to the existing one?  Is this an HNAP validation rue or is it specific for a deployment?

josegar74:
This validation is only relevant when publishing to DOI. For the metadata editor, these validation rules should be disabled to avoid interfering with metadata validation when the metadata is not relevant to the DOI publication.

ianwallen (approved):
It would be nice if the settings could be updated so that by default this schematron would be disabled.  I'm not sure if this is possible so I will approve the PR and we can possibly look into this in the future.

Redge0690 (approved):

josegar74:
@ianwallen the schematron code supports to add `disabled` in the file name (`schematron-rules-datacite.disabled.sch`) do the rule is disabled by default.

Unfortunately, that causes issues with the DOI code as checks for the exact file name `schematron-rules-datacite.sch`.

For now, going to merge the PR, knowing that should be disabled, if required, in the Admin console > Metadata and templates > Validation section

Issue #92, opened 2020-09-24
Thumbnails with description set to  thumbnail_fre and large_thumbnail_fre are not imported correctly


ianwallen:
Metadata which includes French thumbnail (gmd:fileDescription = thumbnail_fre and large_thumbnail_fre) are not imported correctly.

Looks like thumbnail_fre and large_thumbnail_fre is being ignored when updating/adding http:// to the gmd:fileName during import.  

![image](https://user-images.githubusercontent.com/1868233/94183394-bede6d00-fe78-11ea-984c-373bff32f8ff.png)

Before import this is the xml
```
<gmd:graphicOverview>
        <gmd:MD_BrowseGraphic>
          <gmd:fileName>
            <gco:CharacterString>Preview_Image_s.png</gco:CharacterString>
          </gmd:fileName>
          <gmd:fileDescription>
            <gco:CharacterString>thumbnail</gco:CharacterString>
          </gmd:fileDescription>
          <gmd:fileType xmlns:napm="http://www.geconnections.org/nap/napMetadataTools/napXsd/napm" xsi:type="napm:napMD_FileFormatCode_PropertyType" codeList="http://nap.geogratis.gc.ca/metadata/register/registerItemClasses-eng.html#IC_115" codeListValue="RI_716">
            <gco:CharacterString>png; png</gco:CharacterString>
          </gmd:fileType>
        </gmd:MD_BrowseGraphic>
      </gmd:graphicOverview>
      <gmd:graphicOverview>
        <gmd:MD_BrowseGraphic>
          <gmd:fileName>
            <gco:CharacterString>Preview_Image.png</gco:CharacterString>
          </gmd:fileName>
          <gmd:fileDescription>
            <gco:CharacterString>large_thumbnail</gco:CharacterString>
          </gmd:fileDescription>
          <gmd:fileType xmlns:napm="http://www.geconnections.org/nap/napMetadataTools/napXsd/napm" xsi:type="napm:napMD_FileFormatCode_PropertyType" codeList="http://nap.geogratis.gc.ca/metadata/register/registerItemClasses-eng.html#IC_115" codeListValue="RI_716">
            <gco:CharacterString>png; png</gco:CharacterString>
          </gmd:fileType>
        </gmd:MD_BrowseGraphic>
      </gmd:graphicOverview>
      <gmd:graphicOverview>
        <gmd:MD_BrowseGraphic>
          <gmd:fileName>
[...]

PR #284, opened 2022-09-14
Update iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl

wangf1122:
During iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP conversion when importing metadata, the resource name in the thumbnail is missing.
image

Therefore, current schematron validation will fail
image

After the fix, the default name will be added

image

No more validation error on thumbnail part.
image

### DIFF
--- a/src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
+++ b/src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
@@ -120,10 +120,9 @@
   </xsl:template>
 
   <!--Add default LocalisedCharacterString to thumbnail -->
-  <xsl:template match="gmd:fileDescription">
+  <xsl:template match="gmd:fileDescription[gco:CharacterString/text() != ''  and not(gmd:PT_FreeText)]">
     <xsl:copy copy-namespaces="no">
-      <xsl:if test="not(gmd:PT_FreeText)">
-        <gco:CharacterString><xsl:value-of select="./gco:CharacterString/text()"/></gco:CharacterString>
+      <xsl:apply-templates select="node()|@*"/>
         <gmd:PT_FreeText>
           <gmd:textGroup>
             <xsl:choose>
@@ -140,8 +139,6 @@
             </xsl:choose>
           </gmd:textGroup>
         </gmd:PT_FreeText>
-      </xsl:if>
-      <xsl:apply-templates select="node()/gco:CharacterString|@*"/>
     </xsl:copy>
   </xsl:template>
 
--- a/src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
+++ b/src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
@@ -122,23 +122,26 @@
   <!--Add default LocalisedCharacterString to thumbnail -->
   <xsl:template match="gmd:fileDescription">
     <xsl:copy copy-namespaces="no">
-      <gmd:PT_FreeText>
-        <gmd:textGroup>
-          <xsl:choose>
-            <xsl:when test="$mainLanguage='eng'">
-              <gmd:LocalisedCharacterString locale="#fra">
-                <xsl:value-of select="./gco:CharacterString/text()"/>
-              </gmd:LocalisedCharacterString>
-            </xsl:when>
-            <xsl:otherwise>
-              <gmd:LocalisedCharacterString locale="#eng">
-                <xsl:value-of select="/gco:CharacterString"/>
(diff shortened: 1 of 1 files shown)


ianwallen:
Related to #92  but does not really fix the problem

ianwallen:
In some cases the data is as follows.

```
         <gmd:graphicOverview xlink:role="urn:xml:lang:fra-CAN">
         </gmd:graphicOverview>
         <gmd:graphicOverview xlink:role="urn:xml:lang:eng-CAN">
         </gmd:graphicOverview>
```

In these cases, the language is based on the urn:xml:lang - so the English and French will be in separate records.

But your fix seems to be simply adding the missing gmd:PT_FreeText which seems to match the concept used in existing FGP records. 

So what you are proposing seems to be a partial fix but I wonder if the fix should be done on the existing FGP data?

@josegar74, @bo-lu  do you have any comments about the missing gmd:PT_FreeText in FGP data?

bo-lu:
I'm not too sure what is appropriate here as I am overall not too concerned since the thumbnails are not displayed on Open Maps. It is an optional field (nullable) but if it is non-null then both English and French text must be provided. 

As you point out, there are a lot of records with missing values in FGP while the editor now writes null values. We can explore an upgrade of FGP records to include missing gmd:PT_FreeText, but my preference is to have it completed via an SQL query (after a lot of testing).

src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
```diff
@@ -119,6 +119,29 @@
...
+            </xsl:when>
+            <xsl:otherwise>
+              <gmd:LocalisedCharacterString locale="#eng">
+                <xsl:value-of select="/gco:CharacterString"/>
+              </gmd:LocalisedCharacterString>
+            </xsl:otherwise>
+          </xsl:choose>
+        </gmd:textGroup>
+      </gmd:PT_FreeText>
+      <xsl:apply-templates select="node()|@*"/>
```
josegar74, line 125: This copies the `gco:CharacterString` after `gmd:PT_FreeText` causing a xsd error.

Another stuff is that the code doesn't check if the `gmd:PT_FreeText` is already defined with a value.

wangf1122: Can you be more specific about what xsd error or provide a sample test data here? I can look for handle existing gmd:PT_FreeText issue.

I have edited the xsl to have the PT_FreeText added after the gco:CharacterString. I also tested it with another FGP record again and don't see any error during transformation.

src/main/config/conversion/import/iso19139.ca.HNAP.FGP-to-iso19139.ca.HNAP.xsl
```diff
@@ -119,6 +119,32 @@
     </xsl:copy>
   </xsl:template>
 
+  <!--Add default LocalisedCharacterString to thumbnail -->
+  <xsl:template match="gmd:fileDescription">
```
ianwallen: @wangf1122  - According to the following http://www.datypic.com/sc/niem21/e-gmd_fileDescription-1.html, there can be more elements than gco:CharacterString within gmd:fileDescription (although I have not seen any examples of this) - So I think you should use something like the following <xsl:apply-templates select="node()|@*"/> instead of specifically using gco:CharacterString

Also have you considered putting the condition in the template so that it only gets applied where needed? It is odd to have a template that starts with and if statement as that condition could have been placed in the template.

Something like the following (I have not tested)

```
<xsl:template match="gmd:fileDescription[gco:CharacterString/text() != ''  and not(gmd:PT_FreeText)]">
     <xsl:apply-templates select="node()|@*"/>
     <gmd:PT_FreeText>
          <gmd:textGroup>
            <xsl:choose>
              <xsl:when test="$mainLanguage='eng'">
                <gmd:LocalisedCharacterString locale="#fra">
                  <xsl:value-of select="./gco:CharacterString/text()"/>
                </gmd:LocalisedCharacterString>
              </xsl:when>
              <xsl:otherwise>
                <gmd:LocalisedCharacterString locale="#eng">
                  <xsl:value-of select="./gco:CharacterString/text()"/>
                </gmd:LocalisedCharacterString>
              </xsl:otherwise>
            </xsl:choose>
          </gmd:textGroup>
        </gmd:PT_FreeText>
</xsl:template>
```

wangf1122: I adjust your xsl a little. The xsl:copy was missing therefore the gmd:fileDescription won't be copied. After add that in, it's working.

ianwallen (approved):

josegar74 (approved):

PR #295, opened 2022-12-19
config data-gn-onlinesrc-list

wangf1122:
This PR is to remove "Link to a Service" and "Link to a Feature" from the online src drop down list. 

The original view is 
image

here is the new view:

image


wangf1122:
This change will only applies to our end. It's not meant to the community. I will close the PR for now

Issue #285, opened 2022-09-21
Add support for adding new Distribution Formats in editor simple view.


ianwallen:
Currently not possible to add extra Distribution Formats unless using the advanced view. Need to apply changes to allow adding a "+" button to add new format.

This seems to currently exists in the ISO19139 schema.

ianwallen:
Looks ISO 19139 may have added this functionality in the following commit
https://github.com/geonetwork/core-geonetwork/commit/8f52f5f6d785c13dc9e91c7e7a90b085703c31d4 but that is old. So may want to look into more recent code to see how it is applied.

ianwallen:
Note that this may also apply to other fields as well - i.e. contacts.

ianwallen:
I just did a test on https://apps.titellus.net/ for Template for Vector data in ISO19139 (multilingual)  and see the following
![image](https://user-images.githubusercontent.com/1868233/195072484-b5a55739-3639-48d3-b8a5-bcf5dcbfb3bd.png)

Even this seems buggy as it seems to be missing "Distributor contact" next to the second "+"

PR #297, opened 2023-01-06
Add distribution format

wangf1122:
Solution for this issue https://github.com/metadata101/iso19139.ca.HNAP/issues/285
This change gives the "add distribution format" button visible.
image

Upon clicking the button, this section/template/snippet will be duplicated to multiple

image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
@@ -502,6 +502,24 @@
                    or="distributionFormat"/>
           </section>
 
+          <action type="add" name="distributionFormat" or="distributionFormat"
+                  in="/gmd:MD_Metadata/gmd:distributionInfo/gmd:MD_Distribution">
+            <template>
+              <snippet>
+                <gmd:distributionFormat>
+                  <gmd:MD_Format>
+                    <gmd:name>
+                      <gco:CharacterString>DOC</gco:CharacterString>
+                    </gmd:name>
+                    <gmd:version gco:nilReason="missing">
+                      <gco:CharacterString/>
+                    </gmd:version>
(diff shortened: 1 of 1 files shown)


src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
```diff
@@ -502,6 +502,24 @@
...
           </section>
 
+          <action type="add" name="distributionFormat" or="distributionFormat"
+                  in="/gmd:MD_Metadata/gmd:distributionInfo/gmd:MD_Distribution">
+            <template>
+              <snippet>
+                <gmd:distributionFormat>
+                  <gmd:MD_Format>
+                    <gmd:name>
+                      <gco:CharacterString>DOC</gco:CharacterString>
```
ianwallen: Sould we be defaultiong the name to DOC?  It may be better to leave this empty.

wangf1122: yeah. i tried empty string. It's not giving me error but put empty to that field. It was defaulted as DOC when we new a metadata. But I think empty for second distribution format is also ok. I changed the code.

src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
```diff
@@ -509,7 +509,7 @@
                 <gmd:distributionFormat>
                   <gmd:MD_Format>
                     <gmd:name>
-                      <gco:CharacterString>DOC</gco:CharacterString>
+                      <gco:CharacterString></gco:CharacterString>
```
ianwallen: Should be able to use `<gco:CharacterString/>` instead of `<gco:CharacterString></gco:CharacterString>`

PR #300, opened 2023-02-03
Spatial representation type should not be mandatory

### LINKED ISSUE #299: Spatial representation type should not be mandatory
It should be possible to create a record without the Spatial representation type

But it keeps adding the Spatial representation type if it is deleted.

josegar74:
Fixes #299

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
@@ -373,11 +373,6 @@
 
       <xsl:apply-templates select="gmd:aggregationInfo" />
       <xsl:apply-templates select="gmd:spatialRepresentationType" />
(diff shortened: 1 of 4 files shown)


wangf1122:
Edited: I misunderstood it. This field still has no empty option to select from and always defaulted to Vector

image

josegar74:
@wangf1122, if no need of the element, you can remove it now. Previously when removed, was re-added again.

wangf1122:
> @wangf1122, if no need of the element, you can remove it now. Previously when removed, was re-added again.

Yes, It can be removed now as 
image


I also checked shematron validation and no related error complain about the removal. Looking good

ianwallen (approved):

jodygarnett:
See https://schemas.metadata.geo.ca/hnap/2.3.1/HNAP.md#footnote2

PR #303, opened 2023-02-15
Update GC_Open_Licenses.rdf

CarHenkel:
Change on two lines HTTP -> HTTPS.  ECCC DMS Management wants to eliminate all HTTP in XML.  @josegar74


bo-lu:
@CarHenkel This may break validation which occurs on the license value. It will also coordination to the 3.6.x branch and upgrade existing records in the existing catalogues (FGP/DFO/ECCC). Let's follow up with Jose so all the needed changes are coordinated.

bo-lu:
I've outlined some tasks in #304 which need to be completed first so existing flows are preserved.

CarHenkel:
Thank you so much! Let's chat about this in the technical community meeting.  We're hoping to tackle some RDF issues next fiscal. (Never done a pull request before.  Hope I went about it the correct way :) Jérôme is tackling a bulk update of existing records

I've implemented the RDF on our staging server, and it does indeed not internally validate until the list item is reselected by the user. I presume it would then fail FGP validation.
https://iso19139echnap.geocat.live/geonetwork/srv/eng/catalog.search#/home

image

bo-lu:
Definitely, it has been on our issues tracker since 2020! It would be nice to have it fixed this FY. I will do what I can on our side. Cheers!

Issue #304, opened 2023-02-15
Replacing HTTP to HTTPS in the "Open Government License - Canada" in 3.6.x and 3.12.x branch


bo-lu:
Currently, the Open Government Licence website will automatically redirected http calls to the https version of the website. This will change in the future and it is good practice to proactively change from http to https
 
The specific text of the licence is:
> * Open Government Licence - Canada (http://open.canada.ca/en/open-government-licence-canada)
> * Licence du gouvernement ouvert - Canada (http://ouvert.canada.ca/fr/licence-du-gouvernement-ouvert-canada)

Proposed steps:
- [ ] 1.1) Update the HNAP 2.3.1 or 2.4 document to use https on gcpedia. Bo
- [x] 1.2) Coordinate with TBS that the URL in the licence needs to be updated to https on their website. Bo
        Change needed on this file: https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/public/static/licenses.json

- [ ] 2.1) Update the schematron to accept https version of the license. Bo
- [ ] 2.2) Change the default license to https on the metadata creation interface. Anthony
- [ ] 2.3) Create and test a SQL command to change existing entries from http to https on all environments. Anthony
- [ ] 2.4) Coordinate testing of the harvester to Open Maps. Hans-Anthony-Bo

- [ ] 3) FGP will notify metadata contributors that the license will be modified from http to https on a certain date. Update will happen automatically. 

- [ ] 4.1) During off-peak hours, FGP tech will turn off GeoNetwork and execute the SQL command per step 2.3. Anthony
- [ ] 4.2) Check harvesting is working properly.

jodygarnett:
Update: There may be more URLs to update from http --> https / ftp --> ftps also 

Note in HNAP all the schema namesapces are ``http`` this does not have to be ``https`` as it is only a URI. The schema location will change from geogratis to a ``https`` location.

bo-lu:
@anthonyfok I think we can proceed on this work if you have time. Much will depend on if bullet 2.2 can be performed easily.

PR #307, opened 2023-03-01
Update EPSG addresses

zhngamy:
Update the EPSG address because the old http://www.epsg-registry.org/ no longer works.  The new address is https://epsg.io.  

http://www.epsg-registry.org/ address changed to https://epsg.io
http://www.spatialreference.org updated to using https


ianwallen (approved):

ianwallen:
Related to #304 and maybe changes should be back ported to 3.6?

PR #310, opened 2023-03-06
remove gmd:characterSet inflating to utf8

wangf1122:
The characterSet in identification (i.e. gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification/gmd:characterSet)

image

is not mandatory. Bu the inflator always inflate this field to UTF8. This PR is to disable such inflating mechanism.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
@@ -116,11 +116,7 @@
       <xsl:apply-templates select="gmd:fileIdentifier" />
       <xsl:apply-templates select="gmd:language" />
       <xsl:apply-templates select="gmd:characterSet" />
-      <xsl:if test="not(gmd:characterSet)">
-        <gmd:characterSet>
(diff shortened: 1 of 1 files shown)


ianwallen:
gmd:characterSet in gmd:MD_DataIdentification is not mandatory however the main gmd:characterSet off of gmd:MD_Metadata is mandatory
So maybe the condition needs to be modified so that only the mandatory ones have the inflate.

wangf1122:
> gmd:characterSet in gmd:MD_DataIdentification is not mandatory however the main gmd:characterSet off of gmd:MD_Metadata is mandatory So maybe the condition needs to be modified so that only the mandatory ones have the inflate.

I just tested from my local. We can make the XPATH more specific for this inflating logic only for gmd:MD_Metadata/gmd:characterSet. It will not affect the gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification/gmd:characterSet and it's still nullable. I will commit the code shortly

ianwallen:
I have tested on my PC and the current changes don't seem to work.  It cause Double character set selection box for the mandatory one.

wangf1122:
> I have tested on my PC and the current changes don't seem to work. It cause Double character set selection box for the mandatory one.

yes, same can be reproduced here as well. I am gonna take this PR down

PR #311, opened 2023-03-07
Update HNAP Level 1 templates

ianwallen:
Update template setting more values empty so that it can be more generic so that it can be applied across Canada.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/templates/hnaplevel1-english.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/templates/hnaplevel1-english.xml
@@ -36,34 +36,49 @@
                </gmd:textGroup>
             </gmd:PT_FreeText>
          </gmd:organisationName>
+         <gmd:positionName gco:nilReason="missing" xsi:type="gmd:PT_FreeText_PropertyType">
+            <gco:CharacterString />
+         </gmd:positionName>
          <gmd:contactInfo>
             <gmd:CI_Contact>
                <gmd:phone>
-                  <gmd:CI_Telephone/>
+                  <gmd:CI_Telephone>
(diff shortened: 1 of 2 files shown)


src/main/plugin/iso19139.ca.HNAP/templates/hnaplevel1-english.xml
```diff
@@ -73,31 +88,39 @@
...
+                        <gmd:LocalisedCharacterString locale="#fra" />
+                     </gmd:textGroup>
+                  </gmd:PT_FreeText>
+               </gmd:hoursOfService>
             </gmd:CI_Contact>
          </gmd:contactInfo>
          <gmd:role>
             <gmd:CI_RoleCode codeList="http://nap.geogratis.gc.ca/metadata/register/napMetadataRegister.xml#IC_90"
-                             codeListValue="RI_418">author; auteur</gmd:CI_RoleCode>
+                             codeListValue="RI_414">pointOfContact; contact</gmd:CI_RoleCode>
```
jodygarnett, line 123: Point of contact is a sensible default and author is not?

ianwallen: @bo-lu this template was copied from FGP - would you be able to answer @jodygarnett's question?

The only main difference from the FGP template is that we have Department of Fisheries and Ocean in the organization and I'm willing to change it to Canada like FGP template if requested.

ianwallen: @jodygarnett  - FGP's suggestion based on their spreadsheet is the "Point of contact".  I would imaging that most people want to know who to contact to get more information.

ianwallen: I will merge as the template is better than what it was - if changes are required we can always make another PR

PR #312, opened 2023-03-09
Email address format validation

wangf1122:
This feature is to validate for all email formatting based on existing schematron validation logic for schematron-rules-multilingual.sch and schematron-rules-non-multilingual.sch

image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-non-multilingual.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-non-multilingual.sch
@@ -160,6 +160,19 @@
     />
   </xsl:function>
 
+  <!--Check email format-->
+  <xsl:function name="geonet:isEmailAddressFormat" as="xs:boolean">
+    <xsl:param name="inputEmailAdress"/>
+        <xsl:choose>
+            <xsl:when test="matches($inputEmailAdress, '^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9._%+-]+\.[a-zA-Z0-9]+$')">
+                <xsl:value-of select="true()"/>
+            </xsl:when>
+            <xsl:otherwise>
+                <xsl:value-of select="false()"/>
+            </xsl:otherwise>
+        </xsl:choose>
+  </xsl:function>
+
   <!-- =============================================================
   EC schematron rules for multilingual validation in metadata editor:
   ============================================================= -->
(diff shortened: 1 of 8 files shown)


src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-non-multilingual.sch
```diff
@@ -160,6 +160,19 @@
     />
   </xsl:function>
 
+  <!--Check email format-->
+  <xsl:function name="geonet:isEmailAddressFormat" as="xs:boolean">
+    <xsl:param name="inputEmailAdress"/>
+        <xsl:choose>
+            <xsl:when test="matches($inputEmailAdress, '^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9._%+-]+\.[a-zA-Z0-9]+$')">
```
ianwallen: I'm not sure if that is the best email match? I have seen emails like `root@localhost` which would not work here - but also not a valid public email.

It you look at the following there is a full list of emails that are probably not supported.
https://github.com/asaskevich/govalidator/issues/131

And the following seems to be the regex used by html5 email check
https://html.spec.whatwg.org/multipage/input.html#valid-e-mail-address (It matches 12 emails from the list or 22 - while yours matches 5 from the list of 22)

Miles.O'Brian@example.com is a valid email that I have seen similar ones in our department.

I recommend that we follow html5 regex.  I would also include a comment indicating where the regex came from

wangf1122: Sounds good. I did quick test with the supplied RegX in the html5, it works fine. I will update this function.

image

wangf1122: The pattern ^[a-zA-Z0-9.!#$%&'*+\/=?^_`{|}~-]+@[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?(?:\.[a-zA-Z0-9](?:[a-zA-Z0-9-]{0,61}[a-zA-Z0-9])?)*$ is extremely hard to work into xslt. I have to use the Java Util function to implemented and it's working.

ianwallen (approved):

ianwallen:
@wangf1122 

I just tested with empty emails and I see the following. As we already have an error for an empty email - there is no need to also show the invalid email.  - I would expect this to only show the missing required email messages in this case.

image

wangf1122:
I did some extra fix to have the validation with allowEmptyEmail option. So in this case, the shematron validation logic wont be duplicated. Here is the test result

image

josegar74 (approved):
LGTM.

PR #318, opened 2023-03-13
Remove default gmd:characterSet utf8 as this field is not mandatory

wangf1122:
This Character Set is not mandatory

image

So when the user create a metadata from its NHAP template, this should be nullable instead of forcing some default value. If we use UTF8 value, the user will never switch back to null value.


wangf1122:
Duplicated with this PR https://github.com/metadata101/iso19139.ca.HNAP/pull/311/files

I will abandon this PR

Issue #319, opened 2023-03-16
gmd:distributionFormat value seems to be incorrect.


ianwallen:
FGP shows the following which uses the code in the <gco:CharacterString>

```
<gmd:distributionFormat>
            <gmd:MD_Format>
               <gmd:name>
                  <gco:CharacterString>CDED ASCII</gco:CharacterString>
               </gmd:name>
               <gmd:version>
                  <gco:CharacterString>test</gco:CharacterString>
               </gmd:version>
            </gmd:MD_Format>
         </gmd:distributionFormat>
```

		 

Our System shows the following which uses the language value in the <gco:CharacterString>
This is incorrect because if this is what is supposed to be there then we would also require the translated text and it does not have the translated text
		 
		 
		 
```
<gmd:distributionFormat>
            <gmd:MD_Format>
               <gmd:name>
                  <gco:CharacterString>ASCII (CDED)</gco:CharacterString>
               </gmd:name>
               <gmd:version>
                  <gco:CharacterString>test</gco:CharacterString>
               </gmd:version>
            </gmd:MD_Format>
         </gmd:distributionFormat>
```

So there bug seems to be related to the editor/validator that should be using codes to follow FGP
Or both English and French values should exists.

ianwallen:
Another issue with current setup is that <gco:CharacterString> is set based on the main language.  But the value set is based on English.

- create metadata record where the main language is French
- Set distribution format to "ASCII (CDED)"
- Validate to ensure it is correct. - It will fail validation with 

![image](https://user-images.githubusercontent.com/1868233/226126621-fe21906c-cb31-465d-abd0-27bf90381b29.png)

This is because it tries to compare the french value to the english value.

It is better if is use codes instead to be consistent

ianwallen:
According to https://github.com/metadata101/schemas/blob/de-geogratis/hnap/2.3.1/HNAP.md#clause-51131-name
It says that the values need to be 
https://www.gcpedia.gc.ca/wiki/Federal_Geospatial_Platform/Policies_and_Standards/Catalogue/Release/Appendix_B_Guidelines_and_Best_Practices/Guide_to_Harmonized_ISO_19115:2003_NAP/Format?setlang=en&uselang=en
But these are all the English values within the src/main/config/codelist/local/thesauri/theme/GC_Resource_Formats.rdf
Switching to the French does not change the codes.  This seems like a bug in the documentation as it does not make sense to use English values - codes exists and are language independent.

[...]

Issue #314, opened 2023-03-10
Issue with inserting online resource format


ianwallen:
There is a bug when attempting to add an online resource format such as ASCII (CDED) 

![image](https://user-images.githubusercontent.com/1868233/224326043-6f0671a3-59e0-417b-a447-b6a6a525b5f6.png)

When validating, it produces the following error.
![image](https://user-images.githubusercontent.com/1868233/224325916-de4c55e2-cc5a-441d-b07b-e964abe0633f.png)

The reason, we get this error is because the data is store as

![image](https://user-images.githubusercontent.com/1868233/224326674-e261fcf0-1d68-4acd-88fe-079a8a8b6d5a.png)

The validation is validating based on the real code which is CDED ASCII and it identifies that it is not the same as ASCII (CDED) and fails.

There was an attempt to fix this but with PR #259.  However this PR was checking against the English code/French Code - when it should be checking the main code.  

The problem is that the main code is not in the data

1 - Is there a bug in saving the data - i.e should it have the main code stored instead of translated values? This does not seem correct as it would be difficult to program to check all codes.
2 - Should the validation accept Eng code, Fra codes, and main codes in the check?

ianwallen:
@CarHenkel Can you test your environment with using values like `ASCII (CDED)` which has different values for Eng code (`ASCII (CDED)`), Fra code (`ASCII (MNEC)`) and main code `CDED ASCII`

With the change reverted, the validation should fail on that item.

ianwallen:
It seems like there are 2 options.
1	Use bilingual values for  the format – in which case the PR #259 can be re-applied and then we need to update the xslt for importing FGP data so that it update the codes to bilingual values.
2	Update the picker so that it stores the code instead of the translation values.  This seems to be related to https://github.com/metadata101/iso19139.ca.HNAP/issues/95

Not sure what the best approach is for this?

CarHenkel:
> @CarHenkel Can you test your environment with using values like `ASCII (CDED)` which has different values for Eng code (`ASCII (CDED)`), Fra code (`ASCII (MNEC)`) and main code `CDED ASCII`
> 
> With the change reverted, the validation should fail on that item.

Hi @ianwallen 

In ECCC environment:
ASCII (CDED) - Passes Validation
ASCII GRID - Passes Validation
[...]

PR #320, opened 2023-03-19
Update distribution format so that the value stored is the code instead of the English value.

ianwallen:
Update distribution format so that the value stored is the code instead of the English value.

Fixes issue #319

Output after this fix will be distribution format similar the following

```
         <gmd:distributionFormat>
            <gmd:MD_Format>
               <gmd:name>
                  <gco:CharacterString>CDED ASCII</gco:CharacterString>
               </gmd:name>
               <gmd:version>
                  <gco:CharacterString>test</gco:CharacterString>
               </gmd:version>
            </gmd:MD_Format>
         </gmd:distributionFormat>
```

Also, the record will validate using a record with the main language as French.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
@@ -290,8 +290,8 @@
       <entries>
         <xsl:for-each select="$resourceFormatsTh/rdf:RDF/rdf:Description">
           <entry>
-            <code><xsl:value-of select="ns2:prefLabel[@xml:lang='en']" /></code>
-            <label> <xsl:value-of select="ns2:prefLabel[@xml:lang='en']" /></label>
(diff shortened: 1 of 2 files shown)


ianwallen:
Todo: test what happens with old data. - is there any dataloss?

jodygarnett:
Based on discussion this is required for compatibility with FGP; however the HNAP guidance document is a bit inconclusive (indicating name is used but then only linking to english name). 

We have asked @bo-lu clarify FGP intent and update the document accordingly.

ianwallen:
> Based on discussion this is required for compatibility with FGP; however the HNAP guidance document is a bit inconclusive (indicating name is used but then only linking to english name).
> 
> We have asked @bo-lu clarify FGP intent and update the document accordingly.

This also applies to #314

ianwallen:
> Todo: test what happens with old data. - is there any data loss?

Tested with an invalid value (GEOJson) and we can see that the value is not removed so there is no data loss.  It also does not validate as expected.

image

josegar74 (approved):
The change works fine. Most of the values are the same in the code and the English value, but for some they are different, for existing installations can be relevant to apply some data update for these values.

PR #326, opened 2023-04-04
Set Core Subject Thesarurus dates in udpate-fixed-info process to override default behavior in GeoNetwork using the current date

### LINKED ISSUE #308: <gmd:thesaurusName> date is being updated when saving metadata record.
Here is a metadata record that I edited and just saved the changed. You can see that the data associated to the thesaurus was changes to the date in which the record was saved.

image

I believe the date of the thesaurus is not supposed to change when saving data as this is the date related to the thesaurus.

- [ ] Update geonetwork core to support dates (https://github.com/geonetwork/core-geonetwork/pull/6972)
- [ ] Fix issue with template to use new date (https://github.com/metadata101/iso19139.ca.HNAP/pull/327)
- [ ] Need to identify dates to be used in core subject thesaurus - https://github.com/metadata101/iso19139.ca.HNAP/issues/329

josegar74:
Fixes #308

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl
@@ -881,7 +881,7 @@
               </gmd:MD_KeywordTypeCode>
             </gmd:type>
 
-            <xsl:copy-of select="keyword[1]/gmd:thesaurusName" />
+            <xsl:apply-templates select="keyword[1]/gmd:thesaurusName" />
           </gmd:MD_Keywords>
         </gmd:descriptiveKeywords>
       </xsl:for-each>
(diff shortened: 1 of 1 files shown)


src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl
```diff
@@ -955,6 +955,51 @@
     </xsl:copy>
   </xsl:template>
 
+  <!-- Fix the dates for the Core Subject Thesaurus -->
```
ianwallen, line 958: This seems to be a bug specifically to fix "Core Subject Thesaurus" Thesarurus

I would have expected a fix that would not change the Thesarurus date if a value already exists.  Otherwise are we not going to have the same issue with the other thesaurus?

wangf1122:
Sometimes, the thesaurus does have date in the thesaurus file which is not last modified date. We do need to ensure that "good" date is not wiped out. 

We probably need to do something on the Java server side to have a flag stating this is last modified date. Then HNAP xsl can recognize it and avoid the transformation.

ianwallen:
@josegar74 is this PR still required? I believe @wangf1122 's fix in PR https://github.com/metadata101/iso19139.ca.HNAP/pull/327 will resolve the issue.

PR #327, opened 2023-04-05
Use dates from the thesaurus file if exists instead of file system last modified date from thesaurusDate

wangf1122:
This is to fix the issue https://github.com/metadata101/iso19139.ca.HNAP/issues/308

Also depends on this PR https://github.com/geonetwork/core-geonetwork/pull/6972

Once Geonetwork passed isFilesystemDate flag into the HNAP schema. HNAP will be able to parse it or discard it.

### DIFF
--- a/src/main/config/codelist/local/thesauri/theme/GC_Core_Subject.rdf
+++ b/src/main/config/codelist/local/thesauri/theme/GC_Core_Subject.rdf
@@ -15,6 +15,8 @@
       <dc:publisher xml:lang="en">Government of Canada; Library and Archives Canada</dc:publisher>
       <dc:publisher xml:lang="fr">Gouvernement du Canada; Bibliothèque et Archives Canada</dc:publisher>
     <dc:description></dc:description>
+    <dcterms:issued>2004</dcterms:issued>
+    <dcterms:created>2016-07-04</dcterms:created>
   </skos:ConceptScheme>
 
   <skos:Concept rdf:about="http://www.thesaurus.gc.ca/concept/#Abbreviations">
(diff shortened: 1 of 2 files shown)


ianwallen:
Related to #329

josegar74:
Probably relevant to update https://github.com/metadata101/iso19139.ca.HNAP/blob/3.12.x/src/main/config/codelist/local/thesauri/theme/GC_Core_Subject.rdf as part of this pull request with the relevant dates.

wangf1122:
Yes, the thesaurus file itself needs to be updated. I was thinking to do in a separate PR or if you prefer, I can do into this PR. I am also trying to define what date to put into this file.

ianwallen:
@josegar74 

> Probably relevant to update https://github.com/metadata101/iso19139.ca.HNAP/blob/3.12.x/src/main/config/codelist/local/thesauri/theme/GC_Core_Subject.rdf as part of this pull request with the relevant dates.

This is related to the following issue
Related to https://github.com/metadata101/iso19139.ca.HNAP/issues/329

Once the correct date is identified then we can create a new PR to apply the changes.

ianwallen (approved):
It would be nice if the `<gmd:date>` was created by a function so that it would remove a lot of duplicated code however overall - I have tested and it works as expected.

ianwallen:
If we don't want to wait for related issue https://github.com/metadata101/iso19139.ca.HNAP/issues/329 , then we can do as @josegar74 suggested and just use the existing values.  And when issue #329 identifies the correct dates then we can create another PR to apply the correct dates.

```
      <dcterms:issued>2004</dcterms:issued>
      <dcterms:created>2016-07-04</dcterms:created>

```

wangf1122:
I added the default date to the GC_Core thesaurus file. Also group the logics to a single function. Tested and works fine. @ianwallen

PR #328, opened 2023-04-20
Add gmd:dataQualityInfo to template and inflate missing ones

wangf1122:
The current template has no such gmd:dataQualityInfo therefore all metadata created are lack of this section. When the user click on "link to a source dataset"
image

Geonetwork will add such section into the metadata. However it wasn't added correctly. The required gmd:scope is missing according to the schema https://github.com/metadata101/iso19139.ca.HNAP/blob/14ead08ef492220fb5d7a6e8c22022807b941201/src/main/plugin/iso19139.ca.HNAP/schema/gmd/dataQuality.xsd#L505

image

As result, the schema validation will fail 
image

This PR will ensure to inflate existing missing gmd:dataQualityInfo also add it to the template. It is backwards compatible by testing with both existing data and adding new data.

image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
@@ -233,11 +233,39 @@
       <xsl:apply-templates select="gmd:identificationInfo" />
       <xsl:apply-templates select="gmd:contentInfo" />
       <xsl:apply-templates select="gmd:distributionInfo" />
-      <xsl:apply-templates select="gmd:dataQualityInfo" />
       <xsl:apply-templates select="gmd:portrayalCatalogueInfo" />
       <xsl:apply-templates select="gmd:metadataConstraints" />
       <xsl:apply-templates select="gmd:applicationSchemaInfo" />
       <xsl:apply-templates select="gmd:metadataMaintenance" />
+
+      <!-- Inflate gmd:dataQualityInfo and add required gmd:scope -->
+      <xsl:if test="(gmd:dataQualityInfo)">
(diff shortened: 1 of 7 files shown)


wangf1122:
@josegar74 @ianwallen 

I made one more commit to reorder the [gmd:dataQualityInfo] after the gmd:distributionInfo according to this http://www.datypic.com/sc/niem21/t-gmd_MD_Metadata_Type.html

Please take a look

josegar74:
Checking the code the process relies in this process: https://github.com/geonetwork/core-geonetwork/blob/3.12.x/schemas/iso19139/src/main/plugin/iso19139/process/source-add.xsl#L56-L64

I think would be better to fix it instead of updating the inflate process of HNAP.

Also about the templates, I don't think they need to be changed?

wangf1122:
> Checking the code the process relies in this process: https://github.com/geonetwork/core-geonetwork/blob/3.12.x/schemas/iso19139/src/main/plugin/iso19139/process/source-add.xsl#L56-L64
> 
> I think would be better to fix it instead of updating the inflate process of HNAP.
> 
> Also about the templates, I don't think they need to be changed?

Hi @josegar74 

Yes, I tested the code in ISO13139 and it works fine. I created a PR to replace this one

https://github.com/geonetwork/core-geonetwork/pull/7068

wangf1122:
@josegar74 

I created this PR for the source-add.xsl as we discussed the codeList of ISO13139 is different. So we should better created from HNAP schema level. Here is the PR: https://github.com/metadata101/iso19139.ca.HNAP/pull/331

josegar74:
Closing this PR, replaced by #331.

PR #332, opened 2023-05-15
adjust xpath for gmd:MD_Metadata/gmd:contact instead of using general…

wangf1122:
gmd:contact exists not only under gmd:metadata but it can appears MD_MaintenanceInformation as well.

Here is the schema definition. 
https://github.com/metadata101/iso19139.ca.HNAP/blob/6066bd8a1504feebf2291d5945a6428d6c002c27/src/main/plugin/iso19139.ca.HNAP/schema/gmd/maintenance.xsd#L27

Therefore the current schematron validates for any //gmd:contact xpath is causing unnecessary validation errors.
image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
@@ -208,7 +208,7 @@
 
 
     <!-- Contact - Organisation Name -->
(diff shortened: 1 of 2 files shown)


ianwallen:
Please explain why MD_MaintenanceInformation that would have a gmd:contact would not be required to follow the same validation?

wangf1122:
I will close this PR as I found the validation is needed for 

image

PR #334, opened 2023-05-26
disable hierarchyLevel field

wangf1122:
The hierarchyLevel  will be disabled like this:

image

The issue is service level metadata has validation of 

https://github.com/metadata101/iso19139.ca.HNAP/blob/6066bd8a1504feebf2291d5945a6428d6c002c27/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch#L410-L417

If this metadata was created as dataset which has no such //srv:SV_ServiceIdentification , then switch to service in the hierarchy level, there will be validation error happens.

image

It's not recommended to switch the hierarchyLevel, therefore disabling it from edit mode is the best option.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
@@ -823,42 +823,42 @@
   </xsl:template>
 
   <!-- Readonly hierarchyLevel in flat mode - changing the hierarchyLevel may cause issues with the xsd schema validation. -->
-  <xsl:template mode="mode-iso19139" priority="2100" match="//gmd:MD_Metadata/gmd:hierarchyLevel[$isFlatMode and $schema = 'iso19139.ca.HNAP']">
-    <xsl:param name="schema" select="$schema" required="no"/>
-    <xsl:param name="labels" select="$labels" required="no"/>
-    <xsl:param name="overrideLabel" select="''" required="no"/>
+<!--  <xsl:template mode="mode-iso19139" priority="2100" match="//gmd:MD_Metadata/gmd:hierarchyLevel[$isFlatMode and $schema = 'iso19139.ca.HNAP']">-->
+<!--    <xsl:param name="schema" select="$schema" required="no"/>-->
+<!--    <xsl:param name="labels" select="$labels" required="no"/>-->
+<!--    <xsl:param name="overrideLabel" select="''" required="no"/>-->
 
-    <xsl:variable name="xpath" select="gn-fn-metadata:getXPath(.)"/>
-    <xsl:variable name="isoType" select="if (../@gco:isoType) then ../@gco:isoType else ''"/>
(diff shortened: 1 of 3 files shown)


josegar74:
To check if this is  a better way, hiding the service value for datasets: https://github.com/geonetwork/core-geonetwork/blob/376e9e6801a493d3163f60f1461317c142537e1a/schemas/iso19139/src/main/plugin/iso19139/loc/eng/codelists.xml#L1644-L1645

wangf1122:
@josegar74 

The hiding way works as well (see service removed from the regular gmd:MD_ScopeCode). I have commit some code and here is the end result. If you are ok with this approach, I will do some final clean up to remove those temporary comment out code.

image

See this service example:

image

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
```diff
@@ -822,4 +822,43 @@
 
   </xsl:template>
 
+  <!-- Readonly hierarchyLevel in flat mode - changing the hierarchyLevel may cause issues with the xsd schema validation. -->
```
ianwallen: Why do we need this commented code?

src/main/plugin/iso19139.ca.HNAP/loc/eng/codelists.xml
```diff
@@ -2023,14 +2023,14 @@
 			<description>Information applies to a computer program or routine</description>
 		</entry>
 		<!-- - - - - - - - - - - - - - - - - - - - - - - - - -->
-		<entry>
```
ianwallen, line 2026: Why do you need this change?
Won't  this be fixed in the followings lines that you added?

```
<codelist name="gmd:MD_ScopeCode"
            displayIf="/ancestor::node()[name()='gmd:MD_Metadata']/gmd:identificationInfo/srv:SV_ServiceIdentification">
```

src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
wangf1122: This was the old code where we decide to disable it. I left there because I didn't know what direction to move forward. Now we decided to reorganize the codelist, then this code is not needed. So i will delete it

src/main/plugin/iso19139.ca.HNAP/loc/eng/codelists.xml
wangf1122: This change is needed. There will be two code list as what Jose's comment. I am adding one for xPath displayIf="/ancestor::node()[name()='gmd:MD_Metadata']/gmd:identificationInfo/srv:SV_ServiceIdentification"

This code here is the original code list which is used for all other type of records. The service needs to be removed because we dont want the user to select service option. Otherwise, the validation will be broken.

josegar74 (approved):

PR #336, opened 2023-06-08
title in batch edit

wangf1122:
This PR fixed the issue in batch editing. With batch edit the title
image

the title will totally wiped out 

image

Issue 1: The operation name GN_Replace was missing from the api call
Issue 2: There is unwanted line change in the API call value

image

This is what it looks like in GN3.12 with ISO19139
image

Fixing Code refers to:
https://github.com/geonetwork/core-geonetwork/blob/57c116c539be00fdef2d7be61d1f72af4f35b405/schemas/iso19139/src/main/plugin/iso19139/layout/config-editor.xml#L3398-L3402

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
@@ -660,9 +660,9 @@
       <field name="title"
              xpath="gmd:identificationInfo/*/
                       gmd:citation/gmd:CI_Citation/gmd:title"
+             insertMode="gn_replace"
              indexField="title">
-        <template>
-          <![CDATA[<gco:CharacterString xmlns:gco='http://www.isotc211.org/2005/gco'>{{value}}</gco:CharacterString>]]></template>
+        <template><![CDATA[<gco:CharacterString xmlns:gco='http://www.isotc211.org/2005/gco'>{{value}}</gco:CharacterString>]]></template>
       </field>
 
       <!-- Insert a new keyword in the first gmd:descriptiveKeywords found.


ianwallen (approved):

josegar74:
The change works, but note that the batch editor requires more updates to use multilingual values.

ianwallen:
@josegar74 
> The change works, but note that the batch editor requires more updates to use multilingual values.

If batch editing is not ready to be used as it required more update, is there an easy way to restrict access to more administrative type users until it is ready?

josegar74:
@ianwallen 

You could hide it from the editor board page: 

https://github.com/geonetwork/core-geonetwork/blob/3359ad55ee8a98d18c0341c6bb96fd9860350d10/web-ui/src/main/resources/catalog/templates/editor/editorboard.html#L67

Changing `ng-if="user.isEditorOrMore()"` to `ng-if="user.isAdministratorOrMore()"`

And adding that snippet to hide it from the menu:

https://github.com/geonetwork/core-geonetwork/blob/3359ad55ee8a98d18c0341c6bb96fd9860350d10/web-ui/src/main/resources/catalog/templates/top-toolbar-accessible.html#L95

Note that will not protect the backend services, but probably enough to hide it to most users.

ianwallen:
@josegar74, @CarHenkel

> Changing `ng-if="user.isEditorOrMore()"` to `ng-if="user.isAdministratorOrMore()"`

Is this something that would be accepted by other users/clients using HNAP or would this need to be an custom change on our side?  We would prefer to restrict the number of custom changes.

CarHenkel:
@ianwallen @josegar74 We had reported this in ticket #356162.  The conversation with Michel came back with two options.  We didn't want to get rid of the batch edit function as my predecessor used it for large edits (I don't).  The bug fell off the radar.
    
### 1. Michel Gabriel
Staff
Hello,

I would propose not to use this batch functions but use the ones in the admin pages instead.

These batch pages can be found under: admin > tools > batch process (or the link: https://iso19139echnap.geocat.live/geonetwork/srv/eng/admin.console#/tools/batch)

We could also remove the batch options from the Contribute menu.


### 2. Michel Gabriel
Staff
I asked around and it's possible to configure the list of fields available for batch processing.

Would it be an option to remove the title field from this list?


FR language support appears to be in.  I support the idea of restricting Batch editing to admin users if we have to. But if the multilingual problems do not affect FR/EN, I don't see a reason to turf it.  If Ian's solution fixes the title field, all the better.

wangf1122:
@josegar74 @ianwallen @CarHenkel 

I have made a separated PR https://github.com/geonetwork/core-geonetwork/pull/7235 to give configuration to turn off the batch editing to some certain level of users. The PR is probably not backportable to 3.12, if so, then I will make another version for 3.12.

ianwallen:
PR to disable batch editing is now on this PR https://github.com/geonetwork/core-geonetwork/pull/7238

josegar74 (approved):

PR #339, opened 2023-06-13
Remove persisting extent element

zhngamy:
Bug - empty Extent element persists in view:
image

Bug fix removes empty Extent element from Basic view.


src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
```diff
@@ -492,21 +492,17 @@
...
-                 if="count(gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification) > 0">
-          </field>
-
           <field name="temporalRangeSection"
                  xpath="/gmd:MD_Metadata/gmd:identificationInfo/srv:SV_ServiceIdentification/gmd:extent/gmd:EX_Extent/gmd:temporalElement"
+                 del="../../../.."
                  if="count(gmd:MD_Metadata/gmd:identificationInfo/srv:SV_ServiceIdentification) > 0">
           </field>
 
-          <section name="gmd:EX_Extent">
```
ianwallen, line 508: I suspect that this is the same as PR #340 and that you only need to add the `displayIfRecord`

ianwallen: I'm not sure why it was changed to a field - I guessing that this was done so that the X existed so that it can be removed.

Changing it to a field seems to have cause problems with the non-spatial template which shows the temporary extents twice 

image

I tried to use the following to exclude the temporal element duplication but that did not work....
`xpath="/gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification/gmd:extent/gmd:EX_Extent[not(gmd:temporalElement)]"`

I'm thinking that we should be using 
`<section name="gmd:EX_Extent" displayIfRecord="count(gmd:MD_Metadata/gmd:identificationInfo/gmd:MD_DataIdentification/gmd:extent/gmd:EX_Extent) > 0">`

There should be a way to properly remove the section? Just not sure how? This may need more testings....

josegar74:
@ianwallen, @zhngamy lease check this PR https://github.com/metadata101/iso19139.ca.HNAP/pull/358, should cover the requirements:

- The editor section for the geographical extent is displayed only if contains a geographical description and/or an extent.
- Update fixed info process, removes extent elements with no children.

PR #340, opened 2023-06-13
Remove persisting referenceSystemInfo element

zhngamy:
Bug- Empty referenceSystemInfo element persists in view:
image

Bug fix removes referenceSystemInfo element from view when is empty element or all child field sets are removed.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
@@ -206,29 +206,6 @@
 
       <xsl:apply-templates select="gmd:spatialRepresentationInfo" />
       <xsl:apply-templates select="gmd:referenceSystemInfo" />
-      <xsl:if test="not(gmd:referenceSystemInfo)">
-        <xsl:variable name="spatialRepresentationType" select="gmd:identificationInfo/*/gmd:spatialRepresentationType/gmd:MD_SpatialRepresentationTypeCode/@codeListValue" />
-
-        <xsl:if test="$spatialRepresentationType = 'RI_635' or $spatialRepresentationType = 'RI_636' or $spatialRepresentationType = 'RI_638'">
-          <gmd:referenceSystemInfo>
-            <gmd:MD_ReferenceSystem>
-              <gmd:referenceSystemIdentifier>
-                <gmd:RS_Identifier>
-                  <gmd:code>
-                    <gco:CharacterString/>
-                  </gmd:code>
-                  <gmd:codeSpace>
-                    <gco:CharacterString/>
-                  </gmd:codeSpace>
-                  <gmd:version>
-                    <gco:CharacterString/>
-                  </gmd:version>
-                </gmd:RS_Identifier>
-              </gmd:referenceSystemIdentifier>
-            </gmd:MD_ReferenceSystem>
-          </gmd:referenceSystemInfo>
-        </xsl:if>
-      </xsl:if>
 
       <xsl:apply-templates select="gmd:identificationInfo" />
       <xsl:apply-templates select="gmd:contentInfo" />
--- a/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
@@ -206,6 +206,29 @@
 
       <xsl:apply-templates select="gmd:spatialRepresentationInfo" />
       <xsl:apply-templates select="gmd:referenceSystemInfo" />
+      <xsl:if test="not(gmd:referenceSystemInfo)">
+        <xsl:variable name="spatialRepresentationType" select="gmd:identificationInfo/*/gmd:spatialRepresentationType/gmd:MD_SpatialRepresentationTypeCode/@codeListValue" />
+
+        <xsl:if test="$spatialRepresentationType = 'RI_635' or $spatialRepresentationType = 'RI_636' or $spatialRepresentationType = 'RI_638'">
+          <gmd:referenceSystemInfo>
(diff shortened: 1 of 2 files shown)


josegar74:
According to HNAP spec:

> Reference System Information (C)*: Mandatory, if spatialRepresentionType in Data Identification is "vector," "grid" or "tin”.

The existing code reintroduces the element if `spatialRepresentionType` has one of these values. I'm not sure if that should be changed.

You can create a template without the `referenceSystemInfo` section and with a value for `spatialRepresentionType` that is not any of the previous one and the `referenceSystemInfo` section should not be re-added.

zhngamy:
> According to HNAP spec:
> 
> > Reference System Information (C)*: Mandatory, if spatialRepresentionType in Data Identification is "vector," "grid" or "tin”.
> 
> The existing code reintroduces the element if `spatialRepresentionType` has one of these values. I'm not sure if that should be changed.
> 
> You can create a template without the `referenceSystemInfo` section and with a value for `spatialRepresentionType` that is not any of the previous one and the `referenceSystemInfo` section should not be re-added.

Hi Jose, thanks for the comment.
Just to confirm, the referenceSystemInfo section should only be removed if the follow criteria is met:

- Is empty section or all child field sets are removed AND
- spatialRepresentionType in Data Identification does not have any of the following values: "vector," "grid" or "tin”

ianwallen:
@josegar74, @zhngamy 

Before this fix, we are able to remove the spatial representation type and we end up with 

image

Which does not have any option to re-add the option. 

This ends up puzzling the users as there is nothing they can do with this section - there is no add or remove so it is just dead text that has no meaning.

We  want to remove this empty block when there no option available to re-add.

I do agree that if there is a Spatial Representation Type value then we should not allow the users to remove the reference system.  But this should also mean that we cannot end up with a blank section like 
image

josegar74:
@ianwallen that happens because the original code doesn't have this in the section: `displayIfRecord="count(gmd:MD_Metadata/gmd:referenceSystemInfo) > 0">`

From the current fix:

https://github.com/metadata101/iso19139.ca.HNAP/pull/340/files#diff-1e5868865766f75732337da83a7301754e5ea6b774c6e714e022dcc8086ca36eR516

The `del="../../../.."` attribute xpath is incorrect. I think it is not needed in this case, with the current value can delete other elements than `gmd:referenceSystemInfo`.

zhngamy:
@josegar74  I've tested that the del="../../../.." attribute xpath can be removed. I've pushed the changes in the latest commit. Would you please review and approve if you have no other comments? Thanks!

src/main/plugin/iso19139.ca.HNAP/layout/config-editor.xml
```diff
@@ -509,11 +509,30 @@
...
-        <section name="referenceSystemInfo">
-          <field xpath="/gmd:MD_Metadata/gmd:referenceSystemInfo"
-                 in="/gmd:MD_Metadata"
-                 or="referenceSystemInfo"/>
+        <section name="referenceSystemInfo" displayIfRecord="count(gmd:MD_Metadata/gmd:referenceSystemInfo) > 0">
+          <field
+            xpath="/gmd:MD_Metadata/gmd:referenceSystemInfo"
+            if="count(gmd:MD_Metadata/gmd:referenceSystemInfo) > 0">
+          </field>
+          <action type="add" name="referenceSystemInfo" or="referenceSystemInfo"
```
josegar74: This is never going to be displayed due to the `section` attribute `displayIfRecord="count(gmd:MD_Metadata/gmd:referenceSystemInfo) > 0"`

As now the section has the field or the button, I would remove the previous attribute from the section.

zhngamy: @josegar74 Have removed the displayIfRecord section attribute in latest commit.

zhngamy: @josegar74 Would you please review and if there are no further conflicts, approve the PR? Thanks!

ianwallen: @zhngamy 

@josegar74 was simply stating that if you have the `displayIfRecord` then the add option will not work.  He did not ask you to remove `displayIfRecord`

I did a quick test where all I changed was 
`<section name="referenceSystemInfo">`
to 
`<section name="referenceSystemInfo" displayIfRecord="count(gmd:MD_Metadata/gmd:referenceSystemInfo) > 0">`

And it seems to have fixed the issue of showing the empty reference system issue.  I believe that is what we are trying to resolve correct?  If so there is no need to have all the other changes.

Maybe you can test on your end and update the PR accordingly.

zhngamy: Have tested and pushed out the latest changes.

src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
```diff
@@ -206,29 +206,6 @@
 
       <xsl:apply-templates select="gmd:spatialRepresentationInfo" />
       <xsl:apply-templates select="gmd:referenceSystemInfo" />
-      <xsl:if test="not(gmd:referenceSystemInfo)">
```
ianwallen: @zhngamy 
Why do you need to remove this code? It only seems to be applied if there is a reference system info and a spatial resprentation type.

zhngamy: Sorry some items were cached while testing. I've removed it now. Thanks!

ianwallen (approved):

josegar74 (approved):

PR #343, opened 2023-07-03
Add open government thesaurus

ianwallen:
Open government has some extra code lists that are not part of HNAP.  The following can be used to add the Open Government data as keyworks thesaurus values as a workaround to the HNAP limitations.

This PR is to add open government thesaurus for audience, subject, geographic region and place of publication

Source: https://github.com/open-data/ckanext-canada/blob/master/ckanext/canada/schemas/presets.yaml

### DIFF
--- a/src/main/config/codelist/local/thesauri/theme/GC_OG_Audience.rdf
+++ b/src/main/config/codelist/local/thesauri/theme/GC_OG_Audience.rdf
@@ -1,0 +1,214 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<rdf:RDF xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
+    xmlns:ns2="http://www.w3.org/2004/02/skos/core#"
+    xmlns:rdfs="http://www.w3.org/2000/01/rdf-schema#"
+    xmlns:skos="http://www.w3.org/2004/02/skos/core#"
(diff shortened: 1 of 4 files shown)


src/main/config/codelist/local/thesauri/theme/GC_OG_Audience.rdf
```diff
@@ -0,0 +1,214 @@
...
+      <dc:title xml:lang="fr">Audience - gouvernement ouvert - gouvernement du Canada</dc:title>
+      <dc:publisher xml:lang="en">Government of Canada; Library and Archives Canada</dc:publisher>
+      <dc:publisher xml:lang="fr">Gouvernement du Canada; Bibliothèque et Archives Canada</dc:publisher>
+
+      <dc:description></dc:description>
+      <dcterms:issued>2016</dcterms:issued>
+      <dcterms:created>2016-04-05</dcterms:created>
+    </skos:ConceptScheme>
+
+    <rdf:Description rdf:about="http://geonetwork-opensource.org/GC_OG/audience#aboriginal_peoples">
```
josegar74, line 23: @ianwallen it should be fine, as other thesaurus use this kind of prefix, but not sure if would be more appropiate to use a more related prefix to Government of Canada, like the one used in the Core Subject Thesaurus?

In any case, as I think you don't use Anchors for keywords is not something critical.

josegar74 (approved):

PR #346, opened 2023-07-27
Add validation of extents

ianwallen:
Add validation of extents
Also extents should be mandatory when gmd:spatialRepresentationType exists.

In the past some of validation was removed for the extents however I think it was incorrect.  It was remove to support non spatial data however instead of removing it completely, it should have probably just been checked when gmd:spatialRepresentationType exists.

This fixes an issue where records would validate however when publishing to FGP, the record would fail due to extent validation issues. Since FGP required spatialRepresentationType, making the check only enforce the rules when spatialRepresentationType  exists seems to fix the issue with FGP and also still works for non-spatial metadata which does not have spatialRepresentationType.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
@@ -386,6 +386,50 @@
 
       <xsl:apply-templates select="gmd:environmentDescription" />
       <xsl:apply-templates select="gmd:extent" />
+
+      <!-- Add mandatory temporal extent if gmd:spatialRepresentationType exists -->
+      <xsl:if test="(gmd:spatialRepresentationType) and (count(gmd:extent/gmd:EX_Extent/gmd:temporalElement[gmd:EX_TemporalExtent/gmd:extent/gml:TimePeriod]) = 0)">
+        <gmd:extent>
+          <gmd:EX_Extent>
+            <gmd:temporalElement>
+              <gmd:EX_TemporalExtent>
+                <gmd:extent>
+                  <gml:TimePeriod gml:id="{generate-id(.)}">
+                    <gml:beginPosition></gml:beginPosition>
+                    <gml:endPosition></gml:endPosition>
+                  </gml:TimePeriod>
+                </gmd:extent>
+              </gmd:EX_TemporalExtent>
+            </gmd:temporalElement>
+          </gmd:EX_Extent>
+        </gmd:extent>
+      </xsl:if>
+
+      <!-- Add mandatory geographic extent if gmd:spatialRepresentationType exists -->
+      <xsl:if test="(gmd:spatialRepresentationType) and (count(gmd:extent/gmd:EX_Extent/gmd:geographicElement[gmd:EX_GeographicBoundingBox]) = 0)">
+        <gmd:extent>
+          <gmd:EX_Extent>
+            <gmd:geographicElement>
+              <gmd:EX_GeographicBoundingBox>
+                <gmd:westBoundLongitude>
+                  <gco:Decimal></gco:Decimal>
+                </gmd:westBoundLongitude>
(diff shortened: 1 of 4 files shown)


jodygarnett (commented):
This change makes sense to me, but I will need to manually test before approving.

aside: In looking at https://github.com/geonetwork/core-geonetwork/blob/main/web/src/test/java/org/fao/geonet/kernel/schema/SchematronRulesIsoTest.java#L65-L69 it appears we can add integration tests.

jodygarnett:
I am having trouble running this:

> Error happened while calling service
> Failed to compile stylesheet. 1 error detected.

```
Error on line 397 column 61 of inflate-metadata.xsl:
  SXXP0003: Error reported by XML parser: The prefix "gml" for element "gml:TimePeriod" is not bound.
2023-08-09 13:21:59,840 ERROR [geonetwork] - Failed to compile stylesheet. 1 error detected.
javax.xml.transform.TransformerConfigurationException: Failed to compile stylesheet. 1 error detected.
	at net.sf.saxon.PreparedStylesheet.prepare(PreparedStylesheet.java:176)
	at net.sf.saxon.TransformerFactoryImpl.newTemplates(TransformerFactoryImpl.java:139)
```

src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
```diff
@@ -386,6 +386,50 @@
...
       <xsl:apply-templates select="gmd:extent" />
+
+      <!-- Add mandatory temporal extent if gmd:spatialRepresentationType exists -->
+      <xsl:if test="(gmd:spatialRepresentationType) and (count(gmd:extent/gmd:EX_Extent/gmd:temporalElement[gmd:EX_TemporalExtent/gmd:extent/gml:TimePeriod]) = 0)">
+        <gmd:extent>
+          <gmd:EX_Extent>
+            <gmd:temporalElement>
+              <gmd:EX_TemporalExtent>
+                <gmd:extent>
+                  <gml:TimePeriod gml:id="{generate-id(.)}">
```
jodygarnett, line 398: I am encountering a problem here:
> Error on line 397 column 61 of inflate-metadata.xsl:
>  SXXP0003: Error reported by XML parser: The prefix "gml" for element "gml:TimePeriod" is not bound.

jodygarnett (changes requested):
Looks like we need to make sure TimePeriod is defined

jodygarnett:
I can find this commit https://github.com/metadata101/iso19139.ca.HNAP/commit/536c78cc650f810cfe7638400934dabace102bc4#diff-f70a3de7483485eb7a8adc9a5aba59c8378954b840ac025a162669ab962e6fd8 which removed a check for ``gmd:spatialRepresentationType``.

I cannot see any reference to ``gml`` in the file history.

Adding `gml` fixes the issue:
```
xmlns:gml="http://www.opengis.net/gml/3.2"
```

⚠️ Looking at language-defaults.xsl a separate namespace is used for xml and gml320 specificly for TimePeriod:
```
                xmlns:gml="http://www.opengis.net/gml/3.2"
                xmlns:gml320="http://www.opengis.net/gml"
                
                <xsl:for-each select="gml:TimePeriod/gml:beginPosition|gml320:TimePeriod/gml320:beginPosition">
```

jodygarnett:
@ianwallen I have pushed up a small fix and this PR is working for me locally now

jodygarnett (commented):
I have addressed the issue I encountered and the editor now functions.

ianwallen:
> I have addressed the issue I encountered and the editor now functions.

@jodygarnett  It is odd that I did not get that error on my locale testing.  In either case, your change looks good - Thank you.

jodygarnett (approved):
Resorted to manual testing, the infrastructure for running schematron tests occurs very late in the build process and is not available during schema plugin compilation.

PR #351, opened 2023-08-04
improve error message for multilingual field security notes

wangf1122:
This PR is improvement PR to display both English and French valid value for security notes in the validation error message

image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-multilingual.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-multilingual.xml
@@ -41,6 +41,8 @@
   <ResourceDescriptionFormat>Online resource description format is not valid. Valid values are:</ResourceDescriptionFormat>
   <ResourceDescriptionLanguage>Online resource description language is not valid. Should be a comma separated values of ISO-LANG-3 codes</ResourceDescriptionLanguage>
   <SecurityLevel>Security User Note is not valid. Valid values are:</SecurityLevel>
+  <SecurityLevelEn>Security User Note is not valid and required by both languages. Valid English values are: </SecurityLevelEn>
+  <SecurityLevelFr> Valid French values are: </SecurityLevelFr>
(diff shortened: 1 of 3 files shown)


src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-multilingual.xml
```diff
@@ -41,6 +41,8 @@
   <ResourceDescriptionFormat>Online resource description format is not valid. Valid values are:</ResourceDescriptionFormat>
   <ResourceDescriptionLanguage>Online resource description language is not valid. Should be a comma separated values of ISO-LANG-3 codes</ResourceDescriptionLanguage>
   <SecurityLevel>Security User Note is not valid. Valid values are:</SecurityLevel>
+  <SecurityLevelEn>Security User Note is not valid and required by both languages. Valid English values are: </SecurityLevelEn>
```
ianwallen, line 44: Why not stay consistent with all other messages- i.e. OtherConstraintsNote2, Keyword, OrganisationNameThesaurus
And simply put 
`Security User Note is required by both languages`

If you really need to include the options then I would recommend using placeholders.  Using multiple message keys to display a message is messy as different languages can change the order of the sentences causing all sorts of issues.  Also it is not clear by looking at the message the order they will need to be concatenated making hard for translators to get the full context of the messages without looking into the code to see how the message is created.

wangf1122: Yeah, i could change the message itself. As for the sequence of error to display with language tag, I am gonna do some small debug on my localhost

Update: I have relook the validation. What I did here to dissemble the original message to two parts. The original error includes both empty and invalid value case. I tried to separate the logic but this validation block here is very complicated. So it's hard to separate the logic. There for I decide to add empty as part of the error also add French values.

wangf1122:
need another round of work to have "both language are required" check for multilingual field like all other fields. Taking this PR off the shelf and will repost it

PR #353, opened 2023-08-11
Mismatching security code in both language validation, missing lang i…

wangf1122:
Added couple of improvement validations.

1) This field is not mandatory so both lang can be missing, but if the user decide to input one lang and skip another. It will returns error:

image

2) The existing logic still works. It will catch the invalid security code.
image

3) Both lang input are valid, but they are mismatched security code. For example one is Protected A and one is protected B

image


ianwallen:
Did some test and I got errors on the following when I was not expecting any error as Unclassified is not supposed to have protected A/B
image

wangf1122:
@ianwallen 

you can pull and test again. I missed the dollar sign in one of the logic variable so that logic was totally skipped

ianwallen (approved):

josegar74 (approved):
Tested and works fine.

PR #355, opened 2023-08-21
validate missing contact email field to be sync as iso19139.nap

wangf1122:
Due to some legacy data that was attempted to publish to FGP, we have validation error at FGP level that contact email is missing. There is no such validation rule in HNAP. So we brought in the validation to be synchronized with this https://github.com/metadata101/iso19139.ca.HNAP/blob/5a2ca4838d95c7c7cc17bc9ec0eb61275311c4fa/src/main/plugin/iso19139.nap/schematron/schematron-rules-ec-multilingual.sch#L265-L269

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
@@ -173,11 +173,6 @@
   <sch:pattern>
     <sch:title>$loc/strings/Metadata</sch:title>
 
-    <sch:rule context="/gmd:MD_Metadata">
-      <sch:let name="missingContactMail" value="not(gmd:contact/*/gmd:contactInfo/*/gmd:address/gmd:CI_Address/gmd:electronicMailAddress)" />
-      <sch:assert test="not($missingContactMail)">$loc/strings/MissingContactMail</sch:assert>
-    </sch:rule>
-
     <!-- Metadata Standard Name -->
     <sch:rule context="//gmd:metadataStandardName">
 
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
@@ -173,6 +173,11 @@
   <sch:pattern>
     <sch:title>$loc/strings/Metadata</sch:title>
 
+    <sch:rule context="/gmd:MD_Metadata">
+      <sch:let name="missingContactMail" value="not(gmd:contact/*/gmd:contactInfo/*/gmd:address/gmd:CI_Address/gmd:electronicMailAddress)" />
+      <sch:assert test="not($missingContactMail)">$loc/strings/MissingContactMail</sch:assert>
+    </sch:rule>
+
     <!-- Metadata Standard Name -->
     <sch:rule context="//gmd:metadataStandardName">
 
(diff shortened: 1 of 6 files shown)


src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
```diff
@@ -173,6 +173,11 @@
   <sch:pattern>
     <sch:title>$loc/strings/Metadata</sch:title>
 
+    <sch:rule context="/gmd:MD_Metadata">
```
ianwallen: I'm not sure why this is multilingual check?  It is not checking values based on multilingual.  I think it should be in  common.

wangf1122: I have moved this logic into the common validation

ianwallen (approved):

josegar74 (commented):
The xsd is modified for HNAP, to make this element mandatory:

https://github.com/metadata101/iso19139.ca.HNAP/blob/a4dd5e0d6aa240b74bff29879e2a31d0fc610ec4/src/main/plugin/iso19139.ca.HNAP/schema/gmd/citation.xsd#L87

It's ok to have this rule in the schematron, but GeoNetwork should not allow to remove the metadata element in the metadata editor. 

If you create metadata with other tool without the email address, you need to check your tool as it's creating invalid metadata.

Another element that could be relevant to validate is the organisation name, the cardinality is `0..n`, but it's mandatory in HNAP (GeoNetwork editor doesn't allow to remove it either).

ianwallen:
@josegar74 
This error occurred when we imported a record from FGP(HNAP 3.6) and attempted to push it back to FGP(HNAP 3.6).  The record was valid in HNA(3.12)P but when pushed back to FGP(HNAP 3.6) it failed validation because HNAP 3.6 has this validation.  So we simply added the same validation that exists in FGP(HNAP 3.6) so that we are consistent with the checking. 

As you mention, this can only happen if external tool does not enforce this or if the user manually attempts to remove it using the XML editing.

josegar74 (approved):
The change looks fine, but the error is never trigger for me.

If I validate the metadata after removing the element in the XML view, the element gets reintroduced. No validation error anymore.

If I import a metadata without this mail and validate it, without editing it. I get the XSD validation issue.

If it solves your case, looks ok, but I'm curious how do you get this validation error.

ianwallen:
@josegar74 
This seems to be the issue

Your reference to electronicMailAddress 
https://github.com/metadata101/iso19139.ca.HNAP/blob/a4dd5e0d6aa240b74bff29879e2a31d0fc610ec4/src/main/plugin/iso19139.ca.HNAP/schema/gmd/citation.xsd#L87

is only required for the dtd if there is a `<gmd:CI_Address>`

For the record in question that was failing to publish to FGP, there was a contact but no `<gmd:CI_Address>`

```
<gmd:contact xmlns="http://www.isotc211.org/2005/gmd">
<gmd:CI_ResponsibleParty>
<gmd:individualName gco:nilReason="missing">
<gco:CharacterString/>
</gmd:individualName>
<gmd:organisationName xsi:type="gmd:PT_FreeText_PropertyType">
<gco:CharacterString>Government of Canada; Fisheries and Oceans Canada</gco:CharacterString>
<gmd:PT_FreeText>
<gmd:textGroup>
<gmd:LocalisedCharacterString locale="#fra">Gouvernement du Canada; Pêches et Océans Canada</gmd:LocalisedCharacterString>
</gmd:textGroup>
</gmd:PT_FreeText>
</gmd:organisationName>
<gmd:positionName gco:nilReason="missing" xsi:type="gmd:PT_FreeText_PropertyType">
<gco:CharacterString/>
</gmd:positionName>
<gmd:role>
<gmd:CI_RoleCode codeList="https://schemas.metadata.geo.ca/register/napMetadataRegister.xml#IC_90" codeListValue="RI_420">editor; réviseur</gmd:CI_RoleCode>
</gmd:role>
</gmd:CI_ResponsibleParty>
</gmd:contact>

```

So the DTD passed validation however it failed FGP validation due to the missing email that was required.

This PR will fix the issue however I'm not sure if you would prefer a fix in the schema.  Please advise if you would prefer a different fix.

Issue #344, opened 2023-07-05
Permissions on maps and downloads does not seem to work with HNAP


ianwallen:
Steps to reproduce
1. Create 2 users reviewer and registered users.
2. With reviewer create metadata record. Ensure that there is a map and resource to be downloaded.
3. Assign permissions as follows (i.e. disable download and interactive maps)
![image](https://github.com/metadata101/iso19139.ca.HNAP/assets/1868233/27f30ae9-9d41-4dd0-9476-bffee8560c22)
4. Now login using registered users. The user should be able to locate and view the newly created record.
5. The registered user can see the download link and interactive map

Expecting the registered users to not see the download link or interactive map options.

Note: Same test was performed using ISO19139 schema and it worked as expected.

ianwallen:
Closing issue
There is still a related issue where download data is not identified correctly in HNAP. See the following #363

PR #362, opened 2023-11-16
Permissions for edit and maps. 

josegar74:
Related to #344

@ianwallen the configuration added in the pull request was missing.

The problem is that for download, we don't have a specific protocol in HNAP. Maybe we can check if the URL ends with some known formats. What do you think?

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schema-ident.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/schema-ident.xml
@@ -59,7 +59,7 @@
     <filter
       xpath=".//gmd:onLine[*/gmd:protocol/gco:CharacterString = 'WWW:DOWNLOAD-1.0-http--download']"
       ifNotOperation="download"/>
-    <filter xpath="*//gmd:onLine[starts-with(*/gmd:protocol/gco:CharacterString, 'OGC:WMS')]"
+    <filter xpath="*//gmd:onLine[starts-with(*/gmd:protocol/gco:CharacterString, 'OGC:WMS') or starts-with(*/gmd:protocol/gco:CharacterString, 'OGC:WFS') or starts-with(*/gmd:protocol/gco:CharacterString, 'ESRI REST')]"
             ifNotOperation="dynamic"/>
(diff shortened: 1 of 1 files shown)


src/main/plugin/iso19139.ca.HNAP/schema-ident.xml
```diff
@@ -51,4 +51,15 @@
...
 	</autodetect>
+  <filters xmlns:gco="http://www.isotc211.org/2005/gco">
+    <filter xpath=".//*[@gco:nilReason='withheld']"
+            ifNotOperation="editing">
+      <keepMarkedElement gco:nilReason="withheld"/>
+    </filter>
+    <filter
+      xpath=".//gmd:onLine[*/gmd:protocol/gco:CharacterString = 'WWW:DOWNLOAD-1.0-http--download']"
+      ifNotOperation="download"/>
+    <filter xpath="*//gmd:onLine[starts-with(*/gmd:protocol/gco:CharacterString, 'OGC:WMS')]"
```
ianwallen: Would we also need to include protocols that start with "ESRI REST:"?  I'm guessing that this bug exists on iso19139

ianwallen:
> The problem is that for download, we don't have a specific protocol in HNAP. Maybe we can check if the URL ends with some known formats. What do you think?

I think it would be difficult to do it by file extensions.

I think the protocol should be added
image
But I'm not sure on the issues it will cause with FGP so we will need to investigate.

ianwallen:
I created a separate issue to investigate adding the download protocol
https://github.com/metadata101/iso19139.ca.HNAP/issues/363

src/main/plugin/iso19139.ca.HNAP/schema-ident.xml
josegar74: I think you are right, and probably to check also `OGC:WFS`

josegar74:
@ianwallen, updated the permissions for maps with the additional protocols.

ianwallen (approved):

PR #364, opened 2023-11-27
xpath bug fixed.

wangf1122:
The dual creation date was supported for the date validation. But due to some minor bug. It does not retrieve the xpath correctly.

image

I have printed from the local IDE console and found both creation date were selected.
image

So I made the fix

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
@@ -220,8 +220,8 @@
     |//*[@gco:isoType='gmd:MD_DataIdentification']/gmd:citation/gmd:CI_Citation/gmd:date/gmd:CI_Date[gmd:dateType/gmd:CI_DateTypeCode/@codeListValue = 'RI_367']/gmd:date
     |//*[@gco:isoType='srv:SV_ServiceIdentification']/gmd:citation/gmd:CI_Citation/gmd:date/gmd:CI_Date[gmd:dateType/gmd:CI_DateTypeCode/@codeListValue = 'RI_367']/gmd:date">
 
-      <sch:let name="creationDate" value="../../../gmd:date/gmd:CI_Date[gmd:dateType/gmd:CI_DateTypeCode/@codeListValue = 'RI_366']/gmd:date/gco:Date[1]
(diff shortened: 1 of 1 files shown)


ianwallen (approved):

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -220,8 +220,8 @@
     |//*[@gco:isoType='gmd:MD_DataIdentification']/gmd:citation/gmd:CI_Citation/gmd:date/gmd:CI_Date[gmd:dateType/gmd:CI_DateTypeCode/@codeListValue = 'RI_367']/gmd:date
     |//*[@gco:isoType='srv:SV_ServiceIdentification']/gmd:citation/gmd:CI_Citation/gmd:date/gmd:CI_Date[gmd:dateType/gmd:CI_DateTypeCode/@codeListValue = 'RI_367']/gmd:date">
 
-      <sch:let name="creationDate" value="../../../gmd:date/gmd:CI_Date[gmd:dateType/gmd:CI_DateTypeCode/@codeListValue = 'RI_366']/gmd:date/gco:Date[1]
-                                         |../../../gmd:date/gmd:CI_Date[gmd:dateType/gmd:CI_DateTypeCode/@codeListValue = 'RI_366']/gmd:date/gco:DateTime[1]" />
+      <sch:let name="creationDate" value="(../../../gmd:date/gmd:CI_Date[gmd:dateType/gmd:CI_DateTypeCode/@codeListValue = 'RI_366']/gmd:date/gco:Date)[1]
+                                         |(../../../gmd:date/gmd:CI_Date[gmd:dateType/gmd:CI_DateTypeCode/@codeListValue = 'RI_366']/gmd:date/gco:DateTime)[1]" />
```
ianwallen: If you have multiple same date types (publication).  Where there are 2 gco:DateTime and one gco:Date then it looks like you could have the same issue.  I think you need brackets around the full expression.

I have not tested but I was thinking something like the following.
```
<sch:let name="creationDate" value="(../../../gmd:date/gmd:CI_Date[gmd:dateType/gmd:CI_DateTypeCode/@codeListValue = 'RI_366']/gmd:date/gco:Date
                                         | ../../../gmd:date/gmd:CI_Date[gmd:dateType/gmd:CI_DateTypeCode/@codeListValue = 'RI_366']/gmd:date/gco:DateTime)[1]" />
```

wangf1122: ok, i tested and it works as well. See updated code

PR #365, opened 2023-12-01
citation validation

wangf1122:
There was some good validation in https://github.com/geonetwork/core-geonetwork/blob/main/schemas/iso19139/src/main/plugin/iso19139/schematron/schematron-rules-iso.sch

But HNAP schema has no such validation Sschematron in place. I attempt to introduce such Schematron and found they are not so compactable. 

One validation rule I would like to bring in is:
https://github.com/geonetwork/core-geonetwork/blob/bd153b0b9af09db01d0faf7d5cfc6b8708110047/schemas/iso19139/src/main/plugin/iso19139/schematron/schematron-rules-iso.sch#L573-L583

There are many fields (i.e. title) in this citation is mandatory  

image

If we attempt to remove this citation, these mandatory fields are removed as well
image

But the data validates fine:
image

Here is the fixed result:
image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
@@ -153,7 +153,7 @@
   <sch:pattern>
       <sch:title>$loc/strings/requireCitation</sch:title>
       <sch:rule context="//gmd:identificationInfo/*/gmd:citation">
-          <sch:assert test="gmd:CI_Citation | @gco:nilReason">
+          <sch:assert test="gmd:CI_Citation">
               <sch:value-of select="$loc/strings/alert.requiredCitation"/>
           </sch:assert>
       </sch:rule>
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
@@ -149,6 +149,15 @@
     </sch:rule>
   </sch:pattern>
 
+  <!--Citation -->
+  <sch:pattern>
+      <sch:title>$loc/strings/requireCitation</sch:title>
+      <sch:rule context="//gmd:identificationInfo/*/gmd:citation">
+          <sch:assert test="gmd:CI_Citation | @gco:nilReason">
+              <sch:value-of select="$loc/strings/alert.requiredCitation"/>
+          </sch:assert>
+      </sch:rule>
+  </sch:pattern>
 
   <!--- Data Identification pattern -->
   <sch:pattern>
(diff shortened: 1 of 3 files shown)


src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -149,6 +149,15 @@
     </sch:rule>
   </sch:pattern>
 
+  <!--Citation -->
+  <sch:pattern>
+      <sch:title>$loc/strings/requireCitation</sch:title>
+      <sch:rule context="//gmd:identificationInfo/*/gmd:citation">
+          <sch:assert test="gmd:CI_Citation | @gco:nilReason">
```
josegar74: Is it required to check for `@gco:nilReason`?

wangf1122: it was copy from its parent schema iso 19139. I don't see any harm to remain the same logic here even the @gco:nilReason is not presented?

josegar74: Usually `@gco:nilReason` is added when the element doesn't a value. So in this case can be used to indicate no `gmd:CI_Citation`, but not sure if we want this case. I guess we want always that `gmd:citation` contains a `gmd:CI_Citation` so the internal elements are validated also, no?

wangf1122: Yes, its targeted only to gmd:CI_Citation node which is inside of gmd:citation. As for gmd:citation, the schema validation will handle such error. But gmd:CI_Citation is handled by schematron.

I am open to have that nilReason there but keep in mind, this is exactly code from

https://github.com/geonetwork/core-geonetwork/blob/bd153b0b9af09db01d0faf7d5cfc6b8708110047/schemas/iso19139/src/main/plugin/iso19139/schematron/schematron-rules-iso.sch#L573-L583


If we decide to remove it, we want to keep consistent

josegar74: I understand that, but I don't think  ISO19139 requires any field inside `gmd:CI_Citation` to be filled (I can't see any related check in https://github.com/geonetwork/core-geonetwork/blob/bd153b0b9af09db01d0faf7d5cfc6b8708110047/schemas/iso19139/src/main/plugin/iso19139/schematron/schematron-rules-iso.sch), while HNAP does. 

In that sense I don't see a problem the check differs. But also I don't see much sense in an iso19139 metadata with `//gmd:MD_DataIdentification/gmd:citation gco:nilReason="missing"` as that will cause the metadata not having title.

So in summary, doing a PR for main looks ok to me, to remove the `@gco:nilReason` stuff.

wangf1122: I see. Sounds good.

josegar74 (approved):

PR #367, opened 2023-12-14
Add missing rdf:type in GC_OG_Geographic_Region

ianwallen:
Add missing rdf:type in GC_OG_Geographic_Region for PEI

It Without this fix users are not able to select PEI from the thesaurus.


wangf1122 (approved):
It looks good. Seems it was missing at the beginning

PR #369, opened 2023-12-29
check online resource url for duplications

wangf1122:
This metadata contains two duplicated online resource **url+language** combination. This is not allowed when publish to Open Data for none spatial type metadata. Therefore, schema level may need some validation to catch such error.

image


src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -519,6 +521,11 @@
 
       <sch:assert test="$mapWMSCount &lt;= 2">$loc/strings/MapResourcesWMSNumber</sch:assert>
       <sch:assert test="$mapWMSCount = 0 or $mapWMSCount = 2 or $mapWMSCount &gt; 2">$loc/strings/MapResourcesWMS</sch:assert>
+
+      <sch:let name="noneSpatialType" value="not(string(../../gmd:identificationInfo/gmd:MD_DataIdentification/gmd:spatialRepresentationType))"/>
```
ianwallen: Why is this related to the spatial type?  FGP also publishes to Open Data so I would expect that the same issue would exists for spatial.
If FGP has another way to handle this case then maybe we should look into how they handle this?

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -519,6 +521,11 @@
...
+
+      <sch:assert test="not($noneSpatialType) or XslUtilHnap:hasNoDuplicatedOnlineResource($onlineResources)">$loc/strings/hasDuplicatedOnlineResource</sch:assert>
```
ianwallen: I would expect that you should be able to identify the duplicates using xslt without needing to use java?

Here is some quick code created from google search and chat gpt. - But I have not tested it.

```
<xsl:function xmlns:sch="http://purl.oclc.org/dsdl/schematron" name="geonet:hasDuplicateResources" as="xs:boolean">
  <xsl:param name="elements" as="element()*"/>
  
  <xsl:key name="groupKey" match="yourElement" use="concat(gmd:url/gco:CharacterString/text(), '|', gmd:name/gco:CharacterString/text())" />

  <xsl:value-of select="boolean($elements[key('groupKey', concat(gmd:url/gco:CharacterString/text(), '|', gmd:name/gco:CharacterString/text())][2])"/>
</xsl:function>
```

```
<!-- Example usage -->
<sch:let name="hasDuplicates" select="geonet:hasDuplicateResources($onlineResources)"/>
```

ianwallen: I just noticed - the unique resource is based on the url and language - not url and name so you may want to fix that as well.

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
wangf1122: There are many legacy data exists with duplicated url for language purpose. If this validation rule applies to them, how we are going to handle the legacy data? Or we just throw the validation error and let user/editor to correct them?

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
wangf1122: The xsl does not work.

Error at xsl:key on line 82 column 108 of schematron-rules-common.xsl:
  XTSE0010: An xsl:function element must not contain an xsl:key element

XSL function should not contain xsl:key at all according to saxon parser. 

See this is the difficulty of using pure xsl, i dont disagree of using it, but i also tried thousand of way try to find a solution which i couldn't.

ianwallen: Looks like the key needs to be outside the function

The following should work

```
  <xsl:key name="groupKey"
           match="*//gmd:transferOptions/gmd:MD_DigitalTransferOptions/gmd:onLine/gmd:CI_OnlineResource[gmd:linkage/gmd:URL]"
           use="concat(gmd:linkage/gmd:URL/text(), '|', gmd:name/gco:CharacterString/text())"/>

  <xsl:function xmlns:sch="http://purl.oclc.org/dsdl/schematron"
                name="geonet:hasDuplicateResources"
                as="xs:boolean">
    <xsl:param name="elements" as="element()*"/>

    <xsl:variable name="duplicateResources">
      <xsl:for-each-group select="$elements"
                          group-by="concat(gmd:linkage/gmd:URL/text(), '|', gmd:name/gco:CharacterString/text())">

        <xsl:if test="count(current-group()) &gt; 1">
          <xsl:if test="exists(tokenize(current-group()[1]/gmd:description/gco:CharacterString/text(), ';')[3])">
            <xsl:value-of select="concat(tokenize(current-group()[1]/gmd:description/gco:CharacterString/text(), ';')[3], ';')"/>
          </xsl:if>
          <xsl:value-of select="current-group()[1]/gmd:linkage/gmd:URL/text()"/>
        </xsl:if>
      </xsl:for-each-group>
    </xsl:variable>

    <xsl:choose>
      <xsl:when test="exists($duplicateResources)">
        <xsl:message>
          <xsl:value-of select="$duplicateResources"/>
        </xsl:message>
        <xsl:sequence select="true()"/>
      </xsl:when>
      <xsl:otherwise>
        <xsl:sequence select="false()"/>
      </xsl:otherwise>
    </xsl:choose>
  </xsl:function>
```

Note: you may want to change the logic so that if a duplicate is identified then it provides the duplicate url in the error message so that it is easier for the user to know where the duplicate is located.

wangf1122: I tested it. It works good. Thank you. I also made small adjust to have url+language as you suggested in the earlier comment. 

I have commit the latest code and removed all Java related checking logic

wangf1122: I commit another version which will render language,url instead of just boolean.

image

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -57,6 +57,33 @@
     </xsl:for-each>
   </xsl:function>
 
+  <xsl:key name="groupKey"
```
ianwallen: I don't think this is required anymore as I don't see any usage of `groupKey`.
Originally it was used however later I switched to xsl:for-each-group and it does it's own grouping.
Try removing this to see if it still works.

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -57,6 +57,33 @@
...
+    use="concat(gmd:linkage/gmd:URL/text(), '|', gmd:description/gco:CharacterString/text())"/>
+
+  <xsl:function xmlns:sch="http://purl.oclc.org/dsdl/schematron"
+    name="geonet:getDuplicateResources"
+    as="xs:string">
+    <xsl:param name="elements" as="element()*"/>
+
+    <xsl:variable name="duplicateResources">
+      <xsl:for-each-group select="$elements"
+        group-by="concat(gmd:linkage/gmd:URL/text(), '|', gmd:description/gco:CharacterString/text())">
```
ianwallen: `group-by="concat(gmd:linkage/gmd:URL/text(), '|', gmd:description/gco:CharacterString/text())">`
Should be
`group-by="concat(gmd:linkage/gmd:URL/text(), '|', tokenize(current-group()[1]/gmd:description/gco:CharacterString/text(), ';')[3])">`

Otherwise it will group based on the full description but the group by should be on the language.

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -57,6 +57,33 @@
...
+
+        <xsl:if test="count(current-group()) &gt; 1">
+          <xsl:if test="exists(tokenize(current-group()[1]/gmd:description/gco:CharacterString/text(), ';')[3])">
```
ianwallen: The results could return 1 are more records.  If there is only one then we don't need a comma but if there are multiple then there should be a comma to separate the items.

Before this if, you may want to add the following.
```
          <xsl:if test="position() &gt; 1">
            <xsl:text>, </xsl:text>
          </xsl:if>
```

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -57,6 +57,33 @@
...
+            <xsl:value-of select="concat(' ',tokenize(current-group()[1]/gmd:description/gco:CharacterString/text(), ',')[3], ';')"/>
```
ianwallen: You have
`[(), ',')[3], ';')](gco:CharacterString/text(), ',')[3], ';')"/>)`
But I believe the items are tokenized by a` ";"`?  Are you sure this works? Maybe you wanted the following

`gco:CharacterString/text(), ';')[3], ',')"/>`

But I would not suggest using a comma because if there are multiple duplicates separated by a comma then this may get confusing.

src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-common.xml
```diff
@@ -56,7 +56,6 @@
   <ServiceNamespace>srv:SV_ServiceIdentification is misplaced for gmd:MD_ScopeCode: </ServiceNamespace>
   <MissingContactMail>Contact - Electronic mail address is required</MissingContactMail>
 
-  <requireCitation>Citation</requireCitation>
-  <alert.requiredCitation>Data Identification Citation is a required element</alert.requiredCitation>
+  <hasDuplicatedOnlineResource>Duplicated Online Resource not allowed. The duplicated online resource "language, url" is:</hasDuplicatedOnlineResource>
```
ianwallen: There can be one or more duplicate so you should not end with "is:" Also, it is not clear that it is not the full resource that cannot be duplicates, it is the resource URL and language.

So maybe something this may be better.

`Duplicated Online Resource URLs for the same language are not allowed. The following duplicate(s) (language; URL) were identified: `

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
wangf1122: No, you cannot remove this or the language code will be missing

image

wangf1122: ok. I tried and there is no ill effect. This groupKey can be removed.

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
wangf1122: ok, that works. I can add that
image

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
wangf1122: It's a small bug. It doesn't work after I refresh my browser and restarted GN server. Let's just use semi colon instead of any comma here.

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
wangf1122: I have fix all related review and here is the final result looks like:

image

ianwallen (approved):

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -57,6 +57,33 @@
...
 
+
+  <xsl:function xmlns:sch="http://purl.oclc.org/dsdl/schematron"
+    name="geonet:getDuplicateResources"
+    as="xs:string">
+    <xsl:param name="elements" as="element()*"/>
+
+    <xsl:variable name="duplicateResources">
+      <xsl:for-each-group select="$elements"
+        group-by="concat(gmd:linkage/gmd:URL/text(), '|', tokenize(current-group()[1]/gmd:description/gco:CharacterString/text(), ';')[3])">
```
josegar74: I'm not really sure about this code, you are using a `group-by` to group the elements and using the `current-group()` inside the `group-by` clause at the same time. This is not very orthodox, even if apparently seems to work for you.

wangf1122: @josegar74 @ianwallen 

Ian had helped me to update this block of syntax. I tested from localhost and works fine. Please revisit the review.

josegar74 (commented):
The pull request description contains: 

> This metadata contains two duplicated online resource url+name combination.

The code checks the url+language, what it works fine.

I want to confirm that this is the correct check and if so, please update the pull request description.

wangf1122:
> The pull request description contains:
> 
> > This metadata contains two duplicated online resource url+name combination.
> 
> The code checks the url+language, what it works fine.
> 
> I want to confirm that this is the correct check and if so, please update the pull request description.

Sorry about misleading. The description I put is wrong. You are right we are trying to catch url+language. i will change the description

josegar74 (approved):

PR #370, opened 2024-02-13
Add WWW:DOWNLOAD-1.0-http--download

ianwallen:
Add WWW:DOWNLOAD-1.0-http--download for the ability to identify records as downloadable and can be hidden based on permissions.

Also moved HTTPS to the top selection box default to first selection and most urls are HTTPS


josegar74 (commented):
@ianwallen, it looks fine to me to add a specific value for downloads, but maybe we need to check the value with HNAP maintainers.

ianwallen:
Note: HNAP Maintainer has the following in the application

image

It is a free form text field for the protocol with suggested values.  So we can easily add any value we want to the selection.

ianwallen:
@bo-lu - do you have any objections with this change?

ianwallen:
Replacing this PR with 4.4 #461

PR #371, opened 2024-03-08
online resource language and its validation

wangf1122:
This is an unusual scenario. The user select online resource language as eng; fra. But manually change it to eng. But the user may forget change in the French field.
image

So the onlinesrc-add process sends payload 
desc: eng#Web Service;ESRI REST;eng|fra#Service Web;ESRI REST;eng,fra
image

Here is issue 1), the onlinesrc-add only detect the desc string to be end with fra which is not precised. It is better to break the string to check each eng#Web Service;ESRI REST;eng and fra#Service Web;ESRI REST;eng,fra to be end with fra or default to English.

Issue 2) if we set up the online resource by accidentally in issue 1) and add another normal French resource. We will end up have two French resource
image

The current logic only calculate if there are total of two online resources and it does not care if they are in English or French. Therefore it will not return any error at all.
https://github.com/metadata101/iso19139.ca.HNAP/blob/5f7a9d54385f3a55a2af13276229890621b89b55/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch#L526-L530

However in the older version 3.6 which FGP uses, the validation does counting both English and French online resources
https://github.com/metadata101/iso19139.ca.HNAP/blob/5a2ca4838d95c7c7cc17bc9ec0eb61275311c4fa/src/main/plugin/iso19139.nap/schematron/schematron-rules-iso19139.nap-internal.sch#L692-L698

Therefore most recent validation doesn't sync with FGP very well. So data validates as pass but cannot be publish to FGP due to the inconsistent validation. The FGP validation works better and we should bring back the old logic in 3.6.

This error will be captured as:
image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/process/onlinesrc-add.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/process/onlinesrc-add.xsl
@@ -178,12 +178,11 @@ Insert is made in first transferOptions found.
       <xsl:choose>
         <xsl:when test="starts-with($protocol, 'OGC:') and $name != ''">
 
-          <xsl:variable name="descMainLang" select="substring-before($desc, '|')" />
-          <xsl:variable name="descTranslatedLang" select="substring-after($desc, '|')" />
+          <xsl:variable name="descMainLang" select="tokenize(substring-before($desc, '|'), ';')[last()]" />
 
           <xsl:variable name="resourceLang">
             <xsl:choose>
-              <xsl:when test="ends-with($descMainLang, 'fra') and ends-with($descTranslatedLang, 'fra')"><xsl:value-of select="'urn:xml:lang:fra-CAN'"/></xsl:when>
+              <xsl:when test="ends-with($descMainLang, 'fra')"><xsl:value-of select="'urn:xml:lang:fra-CAN'"/></xsl:when>
               <xsl:otherwise><xsl:value-of select="'urn:xml:lang:eng-CAN'"/></xsl:otherwise>
             </xsl:choose>
           </xsl:variable>
@@ -374,12 +373,11 @@ Insert is made in first transferOptions found.
           <gmd:onLine>
             <xsl:variable name="isMapProtocol" select="starts-with($protocol, 'ESRI REST:') or starts-with($protocol, 'OGC:')" />
 
-            <xsl:variable name="descMainLang" select="substring-before($desc, '|')" />
-            <xsl:variable name="descTranslatedLang" select="substring-after($desc, '|')" />
+            <xsl:variable name="descMainLang" select="tokenize(substring-before($desc, '|'), ';')[last()]" />
 
             <xsl:variable name="resourceLang">
               <xsl:choose>
-                <xsl:when test="ends-with($descMainLang, 'fra') and ends-with($descTranslatedLang, 'fra')"><xsl:value-of select="'urn:xml:lang:fra-CAN'"/></xsl:when>
+                <xsl:when test="ends-with($descMainLang, 'fra')"><xsl:value-of select="'urn:xml:lang:fra-CAN'"/></xsl:when>
                 <xsl:otherwise><xsl:value-of select="'urn:xml:lang:eng-CAN'"/></xsl:otherwise>
               </xsl:choose>
             </xsl:variable>
--- a/src/main/plugin/iso19139.ca.HNAP/process/onlinesrc-add.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/process/onlinesrc-add.xsl
@@ -178,9 +178,12 @@ Insert is made in first transferOptions found.
       <xsl:choose>
         <xsl:when test="starts-with($protocol, 'OGC:') and $name != ''">
 
+          <xsl:variable name="descMainLang" select="substring-before($desc, '|')" />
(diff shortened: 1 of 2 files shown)


src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch
```diff
@@ -517,17 +517,20 @@
...
       <sch:let name="uppercase" value="'ABCDEFGHIJKLMNOPQRSTUVWXYZ'" />
 
-      <sch:let name="mapRESTCount" value="count(gmd:transferOptions/gmd:MD_DigitalTransferOptions/gmd:onLine[@xlink:role='urn:xml:lang:eng-CAN' and translate(gmd:CI_OnlineResource/gmd:protocol/gco:CharacterString, $uppercase, $smallcase) = 'esri rest: map service']) +
-                count(gmd:transferOptions/gmd:MD_DigitalTransferOptions/gmd:onLine[@xlink:role='urn:xml:lang:fra-CAN' and translate(gmd:CI_OnlineResource/gmd:protocol/gco:CharacterString, $uppercase, $smallcase) = 'esri rest: map service'])" />
+      <sch:let name="mapRESTCountE" value="count(gmd:transferOptions/gmd:MD_DigitalTransferOptions/gmd:onLine[@xlink:role='urn:xml:lang:eng-CAN' and translate(gmd:CI_OnlineResource/gmd:protocol/gco:CharacterString, $uppercase, $smallcase) = 'esri rest: map service'])" />
+      <sch:let name="mapRESTCountF" value="count(gmd:transferOptions/gmd:MD_DigitalTransferOptions/gmd:onLine[@xlink:role='urn:xml:lang:fra-CAN' and translate(gmd:CI_OnlineResource/gmd:protocol/gco:CharacterString, $uppercase, $smallcase) = 'esri rest: map service'])" />
```
ianwallen, line 521: You seem to be doing language specific checks?  It seems like this would belong in schematron-rules-multilingual.sch

ianwallen: Sorry - I see that the language specific checks for `urn:xml:lang:eng-CAN` and `urn:xml:lang:fra-CAN` were being done before your changes.  So I guess it should be ok

src/main/plugin/iso19139.ca.HNAP/process/onlinesrc-add.xsl
```diff
@@ -371,9 +374,12 @@ Insert is made in first transferOptions found.
           <gmd:onLine>
             <xsl:variable name="isMapProtocol" select="starts-with($protocol, 'ESRI REST:') or starts-with($protocol, 'OGC:')" />
 
+            <xsl:variable name="descMainLang" select="substring-before($desc, '|')" />
+            <xsl:variable name="descTranslatedLang" select="substring-after($desc, '|')" />
+
             <xsl:variable name="resourceLang">
               <xsl:choose>
-                <xsl:when test="ends-with($desc, 'fra')"><xsl:value-of select="'urn:xml:lang:fra-CAN'"/></xsl:when>
+                <xsl:when test="ends-with($descMainLang, 'fra') and ends-with($descTranslatedLang, 'fra')"><xsl:value-of select="'urn:xml:lang:fra-CAN'"/></xsl:when>
```
ianwallen: I would expect that we would only really care about $descMainLang. If $descTranslatedLang is not correct then it is a translation issue.

Otherwise, in the case where someone messes up the translation

`desc: eng#Web Service;ESRI REST;fra|fra#Service Web;ESRI REST;eng`

In this case I think it is better to use `urn:xml:lang:fra-CAN` as identified in the main language than to default to `urn:xml:lang:eng-CAN` because the match failed.

Also we should probably be careful for the following case.
`desc: eng#Web Service;ESRI REST;eng,fra|fra#Service Web;ESRI REST;eng`

In this case it ends with fra but it is not French, it is bilingual. In this case - I'm not sure what it best to use. But it does not seem like French is the best options.  But nor is English.  I wonder if if the last parameter is not "eng" or "fra" if it would be better to default to no value for resourceLang as we don't know.

So in this case, we should not check for 
`substring-before($desc, '|')`
But 
`<xsl:variable name="descMainLang" select="tokenize(substring-before($desc, '|'), ';')[last()]" />`

Then instead of doing 
`ends-with($descMainLang, 'fra')`

you could do `$descMainLang = 'fra'` and you would also need a case for `$descMainLang = 'eng'`

Then the default would be nothing.
 - eng=urn:xml:lang:fra-CAN
 - fra=urn:xml:lang:eng-CAN
 - eng,fra=null
 
 I'm not sure if this would make more sense? - Not sure if the last case of setting it to null will cause issues?

wangf1122: I have made change change to parse the better descMainLang 

But as for the default language setup part, I am hesitate to do it because we haven't studied very well into if the language code of null will impact other things. My personal opinion will just keep the original logic for checking end-with. We could open a new issue or PR to track this specific issue and do some further investigation. But this PR is to mainly fix the validation inconsistent.

wangf1122:
@ianwallen 

4.2.x PR replacement: https://github.com/metadata101/iso19139.ca.HNAP/pull/378

PR #372, opened 2024-03-20
org name validation thesaurus

wangf1122:
The Gc_names were defined in 
https://github.com/metadata101/iso19139.ca.HNAP/blob/5f7a9d54385f3a55a2af13276229890621b89b55/src/main/plugin/iso19139.ca.HNAP/index-fields/default.xsl#L65

and also used and org name 
https://github.com/metadata101/iso19139.ca.HNAP/blob/5f7a9d54385f3a55a2af13276229890621b89b55/src/main/plugin/iso19139.ca.HNAP/index-fields/default.xsl#L392-L402

But the validation was still based on the old thesaurus. We need to adjust the validation rule accordingly.

The change here is to change the suggestion drop down to the user with the better thesaurus

image

The validation rule will apply to these suggested selections. If something is not correct, the error message will still be:

image


ianwallen:
I noticed the error message is not based on the thesaurus

https://github.com/metadata101/iso19139.ca.HNAP/blob/d5798a02a554f46a1ade94029a88e2e5e05e2105/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-non-multilingual.xml#L18

I would expect that the error message should be based on the thesaurus.  This way if someone wants to modify the thesaurus, it will automatically update the error message as well as the drop down list.

I expect it to be similar to the following.
https://github.com/metadata101/iso19139.ca.HNAP/blob/d5798a02a554f46a1ade94029a88e2e5e05e2105/src/main/plugin/iso19139.ca.HNAP/loc/eng/schematron-rules-non-multilingual.xml#L57

It would be nice if we 
- kept GC_Org_Names.rdf for the drop down list default based on value from src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
- Updated error to be base on the value from src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl

This way if someone wants to use a different thesaurus, they would simply modify src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl and it would automatically use the correct thesaurus for the select and validation process.

wangf1122:
@ianwallen 

I am closing this pull request as the direction changed. Making another pull request and we can continue the review over there https://github.com/metadata101/iso19139.ca.HNAP/pull/376

PR #373, opened 2024-03-21
pass index and hash in resource and thumbnail add

wangf1122:
This PR is to coordinate with recent change in Geonetwork ISO 19139 schema change. For online resource and thumbnail add/removal, we need to pass index and hash code to determine the correct resource to add or remove.

For details, please see PR https://github.com/geonetwork/core-geonetwork/pull/7740

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/process/thumbnail-add.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/process/thumbnail-add.xsl
@@ -111,8 +111,6 @@
          [$resourceIdx = '' or position() = xs:integer($resourceIdx)]
          [    ($resourceHash != '' or ($updateKey != '' and normalize-space($updateKey) = concat(
                            */gmd:fileName/gco:CharacterString,
-                           */gmd:fileName/gmd:PT_FreeText/gmd:textGroup/gmd:LocalisedCharacterString[@locale = '#DE'],
-                           */gmd:fileDescription/gmd:PT_FreeText/gmd:textGroup/gmd:LocalisedCharacterString[@locale = '#DE'],
                            */gmd:fileDescription/gco:CharacterString)))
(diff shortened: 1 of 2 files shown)


src/main/plugin/iso19139.ca.HNAP/process/thumbnail-add.xsl
```diff
@@ -90,11 +102,19 @@
...
+  <!-- Note: first part of the match needs to match the xsl:for-each select from extract-relations.xsl in order to get the position() to match -->
+  <!-- The unique identifier is marked with resourceIdx which is the position index and resourceHash which is hash code of the current node (combination of url, resource name, and description) -->
+  <xsl:template
+    priority="2"
+    match="*//gmd:graphicOverview
+         [$resourceIdx = '' or position() = xs:integer($resourceIdx)]
+         [    ($resourceHash != '' or ($updateKey != '' and normalize-space($updateKey) = concat(
+                           */gmd:fileName/gco:CharacterString,
+                           */gmd:fileName/gmd:PT_FreeText/gmd:textGroup/gmd:LocalisedCharacterString[@locale = '#DE'],
+                           */gmd:fileDescription/gmd:PT_FreeText/gmd:textGroup/gmd:LocalisedCharacterString[@locale = '#DE'],
```
josegar74: Is this copied from iso19139? I think we can remove this

wangf1122: yes, looks like it. I don't see any commit about the code, so guess it was long time ago's commit

wangf1122: Sorry, i miss understood it. Yes, it was old code and its kind of related to my change. I removed the DE locale and things should be good. Please review

josegar74 (approved):

PR #374, opened 2024-03-28
Customize search csv export for HNAP

tylerjmchugh:
HNAP Schema is using the iso19139 csv. Added a tpl-csv.xsl customized for HNAP, copied from iso19139's tpl-csv.xsl.

RI codes used in column names are mapped to a meaningful and readable string using codelists.

Without this PR the HNAP headers look like this

image

Values should come from codelists.

```xml
<xsl:variable name="codelists" select="/root/gui/schemas/iso19139.ca.HNAP/codelists"/>
```
Updated some issues related to thesaurus keywords.

Reference system element added so that geoBox coordinates can be used appropriately.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl
@@ -1,0 +1,172 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Copyright (C) 2001-2016 Food and Agriculture Organization of the
+  ~ United Nations (FAO-UN), United Nations World Food Programme (WFP)
+  ~ and United Nations Environment Programme (UNEP)
+  ~
+  ~ This program is free software; you can redistribute it and/or modify
+  ~ it under the terms of the GNU General Public License as published by
+  ~ the Free Software Foundation; either version 2 of the License, or (at
+  ~ your option) any later version.
+  ~
+  ~ This program is distributed in the hope that it will be useful, but
+  ~ WITHOUT ANY WARRANTY; without even the implied warranty of
+  ~ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU
+  ~ General Public License for more details.
+  ~
+  ~ You should have received a copy of the GNU General Public License
+  ~ along with this program; if not, write to the Free Software
+  ~ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301, USA
+  ~
+  ~ Contact: Jeroen Ticheler - FAO - Viale delle Terme di Caracalla 2,
+  ~ Rome - Italy. email: geonetwork@osgeo.org
+  -->
+
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
+                xmlns:gmd="http://www.isotc211.org/2005/gmd"
+                xmlns:gco="http://www.isotc211.org/2005/gco"
+                xmlns:srv="http://www.isotc211.org/2005/srv"
+                xmlns:gmx="http://www.isotc211.org/2005/gmx"
+                xmlns:gml="http://www.opengis.net/gml/3.2"
+                xmlns:gml320="http://www.opengis.net/gml"
+                xmlns:gn="http://www.fao.org/geonetwork"
+                xmlns:gn-fn-core="http://geonetwork-opensource.org/xsl/functions/core"
+                xmlns:gn-fn-iso19139="http://geonetwork-opensource.org/xsl/functions/profiles/iso19139"
+                exclude-result-prefixes="#all"
+                version="2.0">
+
+  <xsl:import href="utility-fn.xsl"/>
+  <xsl:import href="utility-tpl.xsl"/>
+
+  <xsl:template mode="csv" match="gmd:MD_Metadata|*[@gco:isoType='gmd:MD_Metadata']"
+                priority="2">
+    <xsl:variable name="langId" select="gn-fn-iso19139:getLangId(., $lang)"/>
+    <xsl:variable name="info" select="gn:info"/>
+
+    <metadata>
+      <title>
+        <xsl:apply-templates mode="localised"
+                             select="gmd:identificationInfo/*/gmd:citation/*/gmd:title">
+          <xsl:with-param name="langId" select="$langId"/>
+        </xsl:apply-templates>
+      </title>
+      <abstract>
+        <xsl:apply-templates mode="localised" select="gmd:identificationInfo/*/gmd:abstract">
+          <xsl:with-param name="langId" select="$langId"/>
+        </xsl:apply-templates>
+      </abstract>
+
+      <category>
+        <xsl:choose>
+          <xsl:when test="gmd:identificationInfo/srv:SV_ServiceIdentification">service</xsl:when>
+          <xsl:otherwise>dataset</xsl:otherwise>
+        </xsl:choose>
+      </category>
+      <metadatacreationdate>
+        <xsl:value-of select="gmd:dateStamp/*"/>
+      </metadatacreationdate>
+
+      <xsl:for-each select="gmd:identificationInfo/*/gmd:citation/*/gmd:date">
+        <xsl:element name="date-{*/gmd:dateType/*/@codeListValue}">
+          <xsl:value-of select="*/gmd:date/*/text()"/>
+        </xsl:element>
+      </xsl:for-each>
+
+      <xsl:for-each select="gmd:identificationInfo/*/gmd:graphicOverview/*/gmd:fileName">
+        <image>
+          <xsl:value-of select="*/text()"/>
+        </image>
+      </xsl:for-each>
+
+      <!-- All keywords not having thesaurus reference -->
+      <xsl:for-each select="gmd:identificationInfo/*/gmd:descriptiveKeywords/*[not(gmd:thesaurusName)]/gmd:keyword[not(@gco:nilReason)]">
+        <keyword>
+          <xsl:apply-templates mode="localised" select=".">
+            <xsl:with-param name="langId" select="$langId"/>
+          </xsl:apply-templates>
+        </keyword>
+      </xsl:for-each>
+
+      <!-- One column per thesaurus -->
+      <xsl:for-each select="gmd:identificationInfo/*/gmd:descriptiveKeywords/*[gmd:thesaurusName]">
+        <xsl:variable name="thesaurusId" select="normalize-space(gmd:thesaurusName/*/gmd:identifier/*/gmd:code/*/text())"/>
+        <xsl:variable name="thesaurusKey" select="if ($thesaurusId != '') then replace($thesaurusId, '[^a-zA-Z0-9]', '') else position()"/>
+
+        <xsl:for-each select="gmd:keyword[not(@gco:nilReason)]">
+          <xsl:element name="keyword-{$thesaurusKey}">
+            <xsl:apply-templates mode="localised" select=".">
+              <xsl:with-param name="langId" select="$langId"/>
+            </xsl:apply-templates>
+          </xsl:element>
+        </xsl:for-each>
+      </xsl:for-each>
+
+      <!-- One column per thesaurus -->
+      <xsl:for-each select="gmd:identificationInfo/*/gmd:pointOfContact">
+        <xsl:variable name="key" select="*/gmd:role/*/@codeListValue"/>
+
+        <xsl:element name="contact-{$key}">
+          <xsl:apply-templates mode="localised" select="*/gmd:organisationName">
+            <xsl:with-param name="langId" select="$langId"/>
+          </xsl:apply-templates>/
+          <xsl:apply-templates mode="localised" select="*/gmd:contactInfo/*/gmd:address/*/gmd:electronicMailAddress">
+            <xsl:with-param name="langId" select="$langId"/>
+          </xsl:apply-templates>
+        </xsl:element>
+      </xsl:for-each>
+
+      <xsl:for-each select="gmd:identificationInfo/*/gmd:extent/*/gmd:geographicElement/gmd:EX_GeographicBoundingBox">
+        <geoBox>
+          <westBL>
+            <xsl:value-of select="gmd:westBoundLongitude"/>
+          </westBL>
+          <eastBL>
+            <xsl:value-of select="gmd:eastBoundLongitude"/>
+          </eastBL>
+          <southBL>
+            <xsl:value-of select="gmd:southBoundLatitude"/>
+          </southBL>
+          <northBL>
+            <xsl:value-of select="gmd:northBoundLatitude"/>
+          </northBL>
+        </geoBox>
+      </xsl:for-each>
+
+
+      <xsl:for-each select="gmd:identificationInfo/*/*/gmd:MD_Constraints/*">
+        <Constraints>
+          <xsl:copy-of select="."/>
+        </Constraints>
+      </xsl:for-each>
+
+      <xsl:for-each select="gmd:identificationInfo/*/*/gmd:MD_SecurityConstraints/*">
+        <SecurityConstraints>
+          <xsl:copy-of select="."/>
+        </SecurityConstraints>
+      </xsl:for-each>
+
+      <xsl:for-each select="gmd:identificationInfo/*/*/gmd:MD_LegalConstraints/*">
+        <LegalConstraints>
+          <xsl:value-of select="*/text()|*/@codeListValue"/>
+        </LegalConstraints>
+      </xsl:for-each>
+
+
+      <xsl:for-each select="gmd:distributionInfo//gmd:linkage">
+        <link>
+          <xsl:value-of select="*/text()"/>
+        </link>
+      </xsl:for-each>
+      <xsl:for-each select="gmd:distributionInfo//gmd:distributionFormat/*/gmd:name">
+        <format>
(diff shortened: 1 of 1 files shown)


src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl
```diff
@@ -0,0 +1,211 @@
...
+              </keyword-other>
+            </xsl:otherwise>
+          </xsl:choose>
+        </xsl:for-each>
+      </xsl:for-each>
+
+      <!-- All keywords with a valid thesaurus name -->
+      <xsl:for-each select="gmd:identificationInfo/*/gmd:descriptiveKeywords/gmd:MD_Keywords[gmd:thesaurusName/*/gmd:title/(gco:CharacterString|gmx:Anchor)/text() != '']">
+        <xsl:variable name="thesaurusName">
+          <xsl:variable name="translationLanguage" select="gmd:thesaurusName/*/gmd:title/*/*/gmd:LocalisedCharacterString/@locale"/>
```
ianwallen: `gmd:LocalisedCharacterString/@locale` is incorrect as it is a locale code/id and not the actual language.

If you need to get the language then you should use the value withing `$langId` which identifies the main language of the document - There are other xsl that collect the main language and alt language and you can refer to that code.

Code should never check for the id. i.e. `#eng` is just and id and it could be used to identify French language documents (although highly unlikely).  A more accurate example could be French being represented with id #fra or #fre.   In either case if you really need the language then you have to map the id #eng to the official language code to ensure the id is the correct language.   There is multiple xslt that does this so I suggest that you look at other xsl for samples.

src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl
```diff
@@ -0,0 +1,211 @@
...
+          <xsl:choose>
+            <xsl:when test="$translationLanguage = '#eng'">
```
ianwallen: This is incorrect.  If the main language of the document is French then English will be in `gmd:LocalisedCharacterString` and French will be in `gco:CharacterString`
But if the main language of the document is English then it will be reversed - English will be in `gco:CharacterString` and French will be in `gmd:LocalisedCharacterString`

There are other xsl files that correctly extract the values based on the document language - you may want to consider looking into those.

ianwallen:
@tylerjmchugh
Some of the items that you are changing in tpl-csv.xsl may be better fixed in the geonetwork core iso19139 base code as I would expect that the bugs are in that templates as well.

@josegar74 do you see any reason to have HNAP specific implementation for this code? Seems like most of the change can be applied to https://github.com/geonetwork/core-geonetwork/blob/6de7305e63cb08b7976b91e9473e90d750b88483/schemas/iso19139/src/main/plugin/iso19139/layout/tpl-csv.xsl  and maybe even https://github.com/geonetwork/core-geonetwork/blob/6de7305e63cb08b7976b91e9473e90d750b88483/schemas/iso19115-3.2018/src/main/plugin/iso19115-3.2018/layout/tpl-csv.xsl

src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl
tylerjmchugh: https://github.com/metadata101/iso19139.ca.HNAP/blob/859ede0bc2648332cd1044a0e114d57e11d6f2f4/src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl#L121-L128

@ianwallen Is this what you had in mind regarding the localization of thesaurus names? I am now getting the locale id that matches the language code for English and using it within the `localised` template to get the correct translation.

I am working under the assumption that the csv export column headers should always be in English for the consistency of exported data. Let me know if all of the element names should be localized as well.

src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl
```diff
@@ -0,0 +1,210 @@
...
+              </keyword-other>
+            </xsl:otherwise>
+          </xsl:choose>
+        </xsl:for-each>
+      </xsl:for-each>
+
+      <!-- All keywords with a valid thesaurus name -->
+      <xsl:for-each select="gmd:identificationInfo/*/gmd:descriptiveKeywords/gmd:MD_Keywords[gmd:thesaurusName/*/gmd:title/(gco:CharacterString|gmx:Anchor)/text() != '']">
+
+        <!-- Get the localeId matching the english LanguageCode (eng) -->
```
ianwallen: Maybe I'm not understanding what you are trying to do?  Generally you should not need English - generally you need the main language and alternative language

I'm not sure why you are not using `$langId` like all the other calls to `mode="localised"`?

`<xsl:with-param name="langId" select="$langId"/>`

If you need to get the main language or alternative language, then you may want to look at the following

https://github.com/metadata101/iso19139.ca.HNAP/blob/085ee01500873bb768dfef69e0a36814a941944b/src/main/plugin/iso19139.ca.HNAP/index-fields/default.xsl#L38-L63

It gets a main language and an alternate language from the metadata record.

src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl
ianwallen: I don't think it will always be English.
You can test the following for testing.
- English UI - export metadata with English as the main language
- French UI - export metadata with English as the main language
- English UI - export metadata with French as the main language
- French UI - export metadata with French as the main language

tylerjmchugh: Testing results from code from this branch:

- English UI - search result export with English main language -> English column headers, English content
- English UI - search result export with French main language -> English column headers, English content
- French UI - search result export with English main language -> English column headers, French content
- French UI - search result export with French main language -> English column headers, French content (except for mapped category value)

Testing results from prod:

- English UI - search result export with English main language -> English column headers, English content
- English UI - search result export with French main language -> English column headers, English content
- French UI - search result export with English main language -> English column headers, Mixed content (fra remains in English, fre is in French)
- French UI - search result export with French main language -> English column headers, French content (except for mapped category value)

src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl
tylerjmchugh: What I'm trying to do is make the CSV export column headers more readable. And in this context specifically making the thesaurus names in the keyword column headers more descriptive.

So instead of `keyword-geonetworkthesaurusexternalthemeGCCoreSubject` ([already always in English](https://github.com/metadata101/iso19139.ca.HNAP/pull/374#discussion_r1549743144)) derived from the thesaurus identifier code I wanted to use the English translation of the thesaurus title (`keyword-GovernmentofCanadaCoreSubjectThesaurus`).

As all the other column headers are always in English regardless of metadata main language or UI language ([see testing results above](https://github.com/metadata101/iso19139.ca.HNAP/pull/374#discussion_r1549743144)) I thought it made sense to force the use of the English translation of the title. The only thing that is translated right now in prod is the content for each column but this has errors as well when the document uses fra instead of fre (fixed in my code).

I could use `$langId` to translate the thesaurus title into the UI language but then the column headers will be inconsistent as all the other headers are English. I don't think the main language or alt language of the documents would help with this.

ianwallen: By selecting English you are assuming that there is English text in the metadata.
If I create only an French only document then I see the following (also noticed the French char issues which is probably not related to your PR)

image

The logic is taking the language of the first record retrieved - so if I export records with some English only and some French only then the headers will change depending on the first record processed in this case it seems to be French. It is possible that if I process the results differently, then it will use the English field.  This can cause inconsistencies in the results.

image

I have tried other thesaurus and I get the following error
image
So it may be difficult to try to handle all potential data in the keyword title.

Also, if the metadata record contains errors then it will be base the headers in the results based on that record.  Meaning that users could potentially change the headers simply by creating different data.
It seems to be better to hard code the fieldname so that the data does not change the field names?

ianwallen:
I guess there is a reason for HNAP to have it's own version 
HNAP stores codelist as RI_* values and it puts the value as both English and French like the following
image
HNAP has special code which uses the RI_* code to lookup the code list to get the proper value to be displayed.

I believe it is something like this, but you will need to look into it.
`<xsl:apply-templates mode="render-value" select="*/@codeListValue"/>`

src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl
tylerjmchugh: Ok that makes sense, I didn't have enough sample data to catch those cases. Maybe we should continue to use the identifier code for the thesaurus then. Would it be safe to strip out the `geonetworkthesaurusexternaltheme` so `keyword-geonetworkthesaurusexternalthemeGCCoreSubject` -> `keyword-GCCoreSubject`

The French character issues seem to just be with excel not knowing the encoding standard. If imported into excel with UTF-8 it seems fine. Opening the raw text also looks fine.

I did notice another bug where double-quotes are escaped using backslashes instead of another double quote resulting in excel and other software parsing the columns incorrectly. However this would be a change for geonetwork-core. Is there a reason backslashes are used instead of quotes?

[RFC 4180](https://www.ietf.org/rfc/rfc4180.txt#:~:text=If%20double-quotes%20are%20used%20to%20enclose%20fields%2c%20then%20a%20double-quote%0a%20%20%20%20%20%20%20appearing%20inside%20a%20field%20must%20be%20escaped%20by%20preceding%20it%20with%0a%20%20%20%20%20%20%20another%20double%20quote.): If double-quotes are used to enclose fields, then a double-quote appearing inside a field must be escaped by preceding it with another double quote.

ianwallen: Correct, the identifier code is more reliable.

Looking at sample data, it looks something like this.

```
<gmd:identifier>
<gmd:MD_Identifier>
<gmd:code>
<gmx:Anchor xlink:href="http://localhost/geonetwork/srv/api/registries/vocabularies/external.theme.GC_Core_Subject">geonetwork.thesaurus.external.theme.GC_Core_Subject</gmx:Anchor>
</gmd:code>
</gmd:MD_Identifier>
</gmd:identifier>
```

Note that it is possible to upload a thesaurus in themes called myThesaurus and then we can upload another thesaurus in place called myThesaurus.
So they would be external.theme.myThesaurus and external.place.myThesaurus

So the categories(theme/place) need to remain to be unique.

Also same applies for the type of thesaurus as you can see documented here there are different types that can be applied.
https://docs.geonetwork-opensource.org/3.12/administrator-guide/managing-classification-systems/managing-thesaurus/

Regarding geonetwork.thesaurus, keep in mind that geonetwork is just a tool that manages the thesaurus. There is nothing stopping someone from hosting their own thesaurus and using something like 

```
<gmd:identifier>
<gmd:MD_Identifier>
<gmd:code>
<gmx:Anchor xlink:href="http://myhost.com/mythesaurus/external.theme.GC_Core_Subject">external.theme.GC_Core_Subject</gmx:Anchor>
</gmd:code>
</gmd:MD_Identifier>
</gmd:identifier>
```

In this case, if  `geonetwork.thesaurus` is removed then this would conflict with this entry - however, I'm not sure why someone would host their own thesaurus and call it the same name as one hosted on the geonetwork but I guess it could happen.

So keeping the full name is more reliable.  

If anything, there could be a bug with `geonetwork.thesaurus` because if we have multiple geonetwork instances deployed and we create a record where one geonetwork instances points to the thesaurus of another geonetwork instance then it is possible that we have the following 2 entries.

<gmx:Anchor xlink:href="http://localhost/geonetwork1/srv/api/registries/vocabularies/external.theme.GC_Core_Subject">geonetwork.thesaurus.external.theme.GC_Core_Subject</gmx:Anchor>

<gmx:Anchor xlink:href="http://localhost/geonetwork2/srv/api/registries/vocabularies/external.theme.GC_Core_Subject">geonetwork.thesaurus.external.theme.GC_Core_Subject</gmx:Anchor>

However since the record from geonetwork2 would have been added manually, the users who added the entry should have called it `geonetwork2.thesaurus.external.theme.GC_Core_Subject` - so this issue would be tied to a user error.

The other thing - maybe it would be good to have a configuration so that `geonetwork.thesaurus` is configurable. Then, each geonetwork installation could determine what they want as a prefix.  There is already a catalogue name, so maybe, it would be better to use the `catalogue name` instead of hard coding it to `geonetwork.thesaurus`.  I wonder what happens in cases where metadata is harvested from another geonetwork instance which has similar thesaurus name but different content in their thesaurus - This could be a bug.

src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl
```diff
@@ -0,0 +1,199 @@
...
+                  <xsl:with-param name="langId" select="$langId"/>
+                </xsl:apply-templates>
+              </keyword-other>
+            </xsl:otherwise>
+          </xsl:choose>
+        </xsl:for-each>
+      </xsl:for-each>
+
+      <!-- All keywords with a valid thesaurus name -->
+      <xsl:for-each select="gmd:identificationInfo/*/gmd:descriptiveKeywords/*[gmd:thesaurusName/*/gmd:identifier/*/gmd:code/*/text() != '']">
```
ianwallen: It is a little odd to do the loop based on `gmd:thesaurusName/*/gmd:identifier/*/gmd:code/*/text() != '`' for one loop and `gmd:thesaurusName/*/gmd:identifier/*/gmd:code/*/text() = ''` for the other loop. When it has the following

```
<xsl:variable name="thesaurusId" select="normalize-space(gmd:thesaurusName/*/gmd:identifier/*/gmd:code/*/text())"/>
<xsl:variable name="thesaurusKey" select="replace($thesaurusId, '[^a-zA-Z0-9]', '')"/>
```

i.e. lets assume the `gmd:code/*/text() = "@"`
And for another thesauruse the `gmd:code/*/text() = "$"`

In this case the thesaurusId will be '' for both and cause unique conflict errors.

I woule expect 
`[replace(gmd:thesaurusName/*/gmd:identifier/*/gmd:code/*/text(), '[^a-zA-Z0-9]') = '']`
and 
`[replace(gmd:thesaurusName/*/gmd:identifier/*/gmd:code/*/text(), '[^a-zA-Z0-9]') != '']`

It is not critical as I can't see where someone would only use none alpha chars in the value.

src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl
ianwallen: > The French character issues seem to just be with excel not knowing the encoding standard. If imported into excel with UTF-8 it seems fine. Opening the raw text also looks fine.

The issue seems to be fixed in gn 4.  I tested on the following site https://apps.titellus.net/geonetwork and noticed that the file is no longer UTF-8 but ANSI instead.  So we should be able to expect this to be fixed in the next upgrade.

tylerjmchugh:
> I guess there is a reason for HNAP to have it's own version HNAP stores codelist as RI_* values and it puts the value as both English and French like the following image HNAP has special code which uses the RI_* code to lookup the code list to get the proper value to be displayed.
> 
> I believe it is something like this, but you will need to look into it. `<xsl:apply-templates mode="render-value" select="*/@codeListValue"/>`

I'm still trying to figure out how the render-value template relates to the RI Codes.

Right now I am getting the codelist myself:

https://github.com/metadata101/iso19139.ca.HNAP/blob/5015bcfb99e61eea64f5bf6a252e2e2e90dfbba3/src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl#L45

Then I am getting the value element for the entry with a code that matches the RI code. And then I am getting the text before the first semi-colon (english text):

https://github.com/metadata101/iso19139.ca.HNAP/blob/5015bcfb99e61eea64f5bf6a252e2e2e90dfbba3/src/main/plugin/iso19139.ca.HNAP/layout/tpl-csv.xsl#L76

I figured that using just the english is ok for a column key as the codelist values are known and there should always be an english value. I found similar logic in other HNAP files and geonetwork functions.

Is something like this acceptable?

I also saw some examples of the RI code to readable string mapping being done with hardcoded variables but I figure using the codelist is better.

ianwallen:
@tylerjmchugh

> I figured that using just the english is ok for a column key as the codelist values are known and there should always be an english value. I found similar logic in other HNAP files and geonetwork functions.

Using `@codeListValue` would be more accurate however I do see what you mean. There seems to be lots of cases that are using `tokenize()`. 

@josegar74, 
   What is the best way to get the language values?

i.e. in the following 
   
```
 <gmd:dateType>
<gmd:CI_DateTypeCode codeList="http://nap.geogratis.gc.ca/metadata/register/napMetadataRegister.xml#IC_87" codeListValue="RI_366">creation; création</gmd:CI_DateTypeCode>
</gmd:dateType>
```

It seems like getting RI_366 from codelist would be the correct way to get the value.  Parsing `creation; création` seems like it can be buggy and cause issues getting the data based on different languages.

Can you give some guidance.

josegar74:
The codes like `RI_366` are translated in the codelist files (`label` element):

https://github.com/metadata101/iso19139.ca.HNAP/blob/085ee01500873bb768dfef69e0a36814a941944b/src/main/plugin/iso19139.ca.HNAP/loc/eng/codelists.xml#L24

But I think should be quite safe to parse the value `creation; création` as in HNAP for the codelist fields in the XML contains the format `English value;French value`.

josegar74 (approved):
Tested and looks ok

PR #376, opened 2024-04-19
Pass dynamic government name thesaurus to the validation error message

wangf1122:
This is a replacement pull request for this closed one https://github.com/metadata101/iso19139.ca.HNAP/pull/372

The issue is the validation error is hardcoded and not based on the real thesaurus 

image

The fix is about introduce dynamic government thesaurus for couple of errors


ianwallen:
The correct fix would also include modifying the following:
https://github.com/metadata101/iso19139.ca.HNAP/blob/002796e492353b1d481beec4aafedb80e8d84b23/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch#L225

So that it would use the value specified here in the user configuration.
https://github.com/metadata101/iso19139.ca.HNAP/blob/002796e492353b1d481beec4aafedb80e8d84b23/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl#L218

It looks like the correct fix would be to update external.theme.GC_Org_Names so that it comes from a variable similar to `$DefaultMainOrganizationName_en` and then update all code to use that variable.  
https://github.com/metadata101/iso19139.ca.HNAP/blob/002796e492353b1d481beec4aafedb80e8d84b23/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl#L218-L222
Then someone could decide to use external.theme.GC_Org_Names or local.theme.GC_Government_Names simply by changing the configuration.

However this seems like a bigger change and could possibly be done in a separate PR.

For now users would be expected to modify GC_Org_Names to contain the list of organizations that they wish to support. For some this may mean that they will have to overwrite GC_Org_Names with GC_Government_Names.

And I also noticed that there seems to be a potential bug the indexing that also does not use the value defined in the configuration.
https://github.com/metadata101/iso19139.ca.HNAP/blob/002796e492353b1d481beec4aafedb80e8d84b23/src/main/plugin/iso19139.ca.HNAP/index-fields/default.xsl#L65
This should currently no cause any issue because all values from GC_Org_Names are in GC_Government_Names. But if ever they were different then this could also cause issues.

For now I believe you changes are good as it ensure that the thesaurus for picking the organization validates against the same thesaurus in the validation.

wangf1122:
Cherry picked this into higher version 4.2.x pull request https://github.com/metadata101/iso19139.ca.HNAP/pull/377

PR #377, opened 2024-04-25
Pass dynamic government name thesaurus to the validation error message

wangf1122:
This is a replacement pull request for this closed one #372 #376

The issue is the validation error is hardcoded and not based on the real thesaurus 

image

The fix is about introduce dynamic government thesaurus for couple of errors


jodygarnett (approved):
I *really* like this change; thank you.

josegar74 (approved):

PR #378, opened 2024-04-25
online resource language and its validation

wangf1122:
This replaces PR #371

This is an unusual scenario. The user select online resource language as eng; fra. But manually change it to eng. But the user may forget change in the French field.
image

So the onlinesrc-add process sends payload 
desc: eng#Web Service;ESRI REST;eng|fra#Service Web;ESRI REST;eng,fra
image

Here is issue 1), the onlinesrc-add only detect the desc string to be end with fra which is not precised. It is better to break the string to check each eng#Web Service;ESRI REST;eng and fra#Service Web;ESRI REST;eng,fra to be end with fra or default to English.

Issue 2) if we set up the online resource by accidentally in issue 1) and add another normal French resource. We will end up have two French resource
image

The current logic only calculate if there are total of two online resources and it does not care if they are in English or French. Therefore it will not return any error at all.
https://github.com/metadata101/iso19139.ca.HNAP/blob/5f7a9d54385f3a55a2af13276229890621b89b55/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-common.sch#L526-L530

However in the older version 3.6 which FGP uses, the validation does counting both English and French online resources
https://github.com/metadata101/iso19139.ca.HNAP/blob/5a2ca4838d95c7c7cc17bc9ec0eb61275311c4fa/src/main/plugin/iso19139.nap/schematron/schematron-rules-iso19139.nap-internal.sch#L692-L698

Therefore most recent validation doesn't sync with FGP very well. So data validates as pass but cannot be publish to FGP due to the inconsistent validation. The FGP validation works better and we should bring back the old logic in 3.6.

This error will be captured as:
image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/process/onlinesrc-add.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/process/onlinesrc-add.xsl
@@ -230,9 +230,11 @@ Note: It assumes that it will be adding new items in
       <xsl:choose>
         <xsl:when test="starts-with($protocol, 'OGC:') and $name != ''">
 
+          <xsl:variable name="descMainLang" select="tokenize(substring-before($desc, '|'), ';')[last()]" />
+
           <xsl:variable name="resourceLang">
             <xsl:choose>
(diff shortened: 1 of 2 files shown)


ianwallen:
Issue was addressed in #426

PR #380, opened 2024-05-02
Searching index with userNote

wangf1122:
This is for searching indexing with

image

This change has dependency with this PR https://github.com/geonetwork/core-geonetwork/pull/8004 in Geonetwork latest version


davidblasby:
I will merge this and https://github.com/geonetwork/core-geonetwork/pull/8004

I've tested it (for searching for something like "Protected B" and ensuring it is in the Elastic Index document).

I will wait a day or 2 incase anyone has any feedback.
This has also been reviewed by Jose.

josegar74 (approved):

PR #382, opened 2024-05-14
Search CSV Export - Get codelists file directly instead of using "/root/gui/schemas/..."

tylerjmchugh:
Following GeoNetwork 4 upgrade, `tpl-csv` cannot access the codelists. This PR changes `tpl-csv` to use the codelists file directly as `/root/gui/schemas/iso19139/codelists` is inaccessible.

Without this PR csv search exports are unable to transform `RI_*` codes into string values resulting in the output of empty strings.

For example this results in all dates grouped together under "date-" instead of "date-publication, date-creation, ...".

image


josegar74 (approved):
Tested, works fine.

PR #383, opened 2024-05-14
Replace RI codes in security and legal constraints with english string

tylerjmchugh:
When exporting search results as CSV, RI_* codes are shown in the field values.

This PR is to replace the RI_* codes and values output in the csv for the security and legal constraints. For consistency with headers the readable English value string should be used.

Without this PR RI_* codes are not formatted in a consistent way between headers and field values.

Ex. A license in LegalConstraints displays as "RI_606 license; licence" instead of "license"

image


josegar74 (approved):
I haven't tested, but the code changes look fine.

PR #387, opened 2024-06-07
Update Open Government Licence - Canada to use https

josegar74:


ianwallen (approved):

josegar74:
@ianwallen apparently this change is not validated with FGP, that still requires the old value. Have you got issues at DFO with this change?

ianwallen:
As the changes were not merged - we never tested.  But even it is an issue, using HTTPS is recommended and we have an XSLT that we run the changes thought before sending it to FGP.  So we can easily put them back to HTTP for FGP purposes if needed until FGP fixes the issue.

Issue #315, opened 2023-03-10
None mandatory Linkage field has red asterisk mark


wangf1122:
Some none mandatory fields for example linkage has the red astrix mark in front

![image](https://user-images.githubusercontent.com/74916635/224448430-32fae8db-05df-4219-a8c7-3d7824d59605.png)

The label for this field has no such condition as in this code.
https://github.com/metadata101/iso19139.ca.HNAP/blob/33a7855cb24d72425337563d03d13738af713e52/src/main/plugin/iso19139.ca.HNAP/loc/eng/labels.xml#L1298-L1303

I found the logic to place such required mark is probably related to the form builder and its parentEditInfo

https://github.com/geonetwork/core-geonetwork/blob/2731e560fd9e3a8cea16047082cf557d3c0978c5/web/src/main/webapp/xslt/ui-metadata/form-builder.xsl#L116-L120

And this parentEditInfo was passed from HNAP layout.xsl
https://github.com/metadata101/iso19139.ca.HNAP/blob/33a7855cb24d72425337563d03d13738af713e52/src/main/plugin/iso19139.ca.HNAP/layout/layout.xsl#L95-L96

So far, this is my investigation. This gmd:linkage is just one of them. There are couple of other fields are forced to set the gn_required (red astrix ) without stating the condition as mandatory.

josegar74:
The "problem" is that the code check is using also the cardinality of the element, that indeed is mandatory to be present in the XML:

http://www.datypic.com/sc/niem21/t-gmd_CI_OnlineResource_Type.html

But that doesn't really imply that the element must contain a value. 

We could try to remove that check, and verify how affects other elements as probably some will require to add the condition as mandatory to be displayed as currently.

ianwallen:
Isn't the real problem that it should be based on  gmd:onlineResource which is not mandatory according to the following:

http://www.datypic.com/sc/niem21/e-gmd_CI_Contact.html

And only if a the gmd:onlineResource is added then the linkage is required.

josegar74:
`gmd:linkage` is mandatory in http://www.datypic.com/sc/niem21/e-gmd_CI_OnlineResource.html

By default, `gmd:onlineResource` is not added to contacts, unless you have that in your template, or add it in the advanced view, which adds the following:

```
 <gmd:onlineResource>
    <gmd:CI_OnlineResource>
        <gmd:linkage>
            <gmd:URL></gmd:URL>
        </gmd:linkage>
[...]

PR #391, opened 2024-08-06
update required field and not force render element to check with xsd schema

wangf1122:
As the discussion from https://github.com/metadata101/iso19139.ca.HNAP/issues/315

The label.xml will updated with the required field. When it goes to rendering the element, the template should have an option not check the xsd schema. So there is an ongoing change for that in Geonetwork main branch https://github.com/geonetwork/core-geonetwork/pull/8294


wangf1122:
Replaced by https://github.com/metadata101/iso19139.ca.HNAP/pull/405

PR #394, opened 2024-08-13
dynamic msg for online resource validation

xiechangning20:
This is updated version of #392
Which only addresses the dynamic error message for content type of online resource validation

The issue is that content-type validation error should fetch the data from thesaurus, similar to resource format
The current msg
image

Msg after the changes
image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
@@ -878,8 +878,8 @@
 
       <sch:let name="resourceContentTypesListMain" value="geonet:resourceContentTypesList($thesaurusDir,$mainLanguage2char)"/>
       <sch:let name="resourceContentTypesListAlt" value="geonet:resourceContentTypesList($thesaurusDir,$altLanguage2char)"/>
-      <sch:let name="locMsgCtMain" value="geonet:prependLocaleMessage(geonet:appendLocaleMessage($loc/strings/ResourceDescriptionContentType, $resourceContentTypesListMain),  concat(gmd:CI_OnlineResource/gmd:linkage/gmd:URL, ' : '))"/>
-      <sch:let name="locMsgCtAlt" value="geonet:prependLocaleMessage(geonet:appendLocaleMessage($loc/strings/ResourceDescriptionContentType, $resourceContentTypesListAlt),  concat(gmd:CI_OnlineResource/gmd:linkage/gmd:URL, ' : '))"/>
+      <sch:let name="locMsgCtMain" value="geonet:prependLocaleMessage(geonet:appendLocaleMessage($loc/strings/*[name() = concat('ResourceDescriptionContentType', $mainLanguageText)], $resourceContentTypesListMain),  concat(gmd:CI_OnlineResource/gmd:linkage/gmd:URL, ' : '))"/>
+      <sch:let name="locMsgCtAlt" value="geonet:prependLocaleMessage(geonet:appendLocaleMessage($loc/strings/*[name() = concat('ResourceDescriptionContentType', $altLanguageText)], $resourceContentTypesListAlt),  concat(gmd:CI_OnlineResource/gmd:linkage/gmd:URL, ' : '))"/>
 
       <sch:assert test="$contentType = document(concat('file:///', replace(concat($thesaurusDir, '/external/thesauri/theme/GC_Resource_ContentTypes.rdf'), '\\', '/')))//rdf:Description/ns2:prefLabel[@xml:lang=$mainLanguage2char]">$locMsgCtMain</sch:assert>
       <sch:assert test="$contentTypeTranslated = document(concat('file:///', replace(concat($thesaurusDir, '/external/thesauri/theme/GC_Resource_ContentTypes.rdf'), '\\', '/')))//rdf:Description/ns2:prefLabel[@xml:lang=$altLanguage2char]">$locMsgCtAlt</sch:assert>
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
@@ -877,8 +877,7 @@
               ('eng', 'fra', 'spa', 'zxx'))"/>
 
       <sch:let name="resourceContentTypesList" select="geonet:resourceContentTypesList($thesaurusDir,$altLanguage2char)"/>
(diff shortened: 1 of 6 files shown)


src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
```diff
@@ -859,7 +876,8 @@
       <sch:let name="languageTranslated_present" value="geonet:values-in($languageTranslated,
               ('eng', 'fra', 'spa', 'zxx'))"/>
 
-      <sch:let name="locMsgCt" value="geonet:prependLocaleMessage($loc/strings/ResourceDescriptionContentType, concat(gmd:CI_OnlineResource/gmd:linkage/gmd:URL, ' : '))" />
+      <sch:let name="resourceContentTypesList" value="geonet:resourceContentTypesList($thesaurusDir,$altLanguage2char)"/>
+      <sch:let name="locMsgCt" value="geonet:prependLocaleMessage(geonet:appendLocaleMessage($loc/strings/ResourceDescriptionContentType, $resourceContentTypesList),  concat(gmd:CI_OnlineResource/gmd:linkage/gmd:URL, ' : '))"/>
 
       <sch:assert test="($contentType = 'Web Service' or $contentType = 'Service Web' or
```
tylerjmchugh, line 882: If the message is generated using thesaurus values should this be checked using the thesaurus values as well rather than hardcoded? This way additional custom content types can be added.

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
```diff
@@ -859,7 +876,8 @@
...
               $contentType = 'Dataset' or $contentType = 'Données' or
```
tylerjmchugh: If the message is generated using thesaurus values should this assert be checked using the thesaurus values as well rather than hardcoded? This way additional custom content types can be added.

tylerjmchugh (approved):
Other than my suggestion with the asserts, everything looks good. Tested locally and it works as expected

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
```diff
@@ -859,7 +876,8 @@
       <sch:let name="languageTranslated_present" value="geonet:values-in($languageTranslated,
               ('eng', 'fra', 'spa', 'zxx'))"/>
 
-      <sch:let name="locMsgCt" value="geonet:prependLocaleMessage($loc/strings/ResourceDescriptionContentType, concat(gmd:CI_OnlineResource/gmd:linkage/gmd:URL, ' : '))" />
+      <sch:let name="resourceContentTypesList" value="geonet:resourceContentTypesList($thesaurusDir,$altLanguage2char)"/>
```
josegar74: For a metadata with main language English, the values displayed are always in French, doesn't matter if the content type invalid value is in English or French:

> Online resource description content type is not valid. Valid values are: API, Application, Données, Document de soutien, Service Web

I think the correct way is to retrieve the message for English and French and depending which value fails, use one or another.

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
xiechangning20: Updated to use thesaurus values in the assertion

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
xiechangning20: Updated based on suggestion
Added message for both English and French Validation
Also updated similar logic in non-multilingual.sch to use thesaurus values in the assertion

src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-non-multilingual.sch
```diff
@@ -504,12 +522,10 @@
...
 
-      <sch:let name="locMsgCt" value="geonet:prependLocaleMessage($loc/strings/ResourceDescriptionContentType, concat(gmd:CI_OnlineResource/gmd:linkage/gmd:URL, ' : '))" />
+      <sch:let name="resourceContentTypesListMain" value="geonet:resourceContentTypesList($thesaurusDir,$mainLanguage2char)"/>
+      <sch:let name="locMsgCt" value="geonet:prependLocaleMessage(geonet:appendLocaleMessage($loc/strings/ResourceDescriptionContentType, $resourceContentTypesListMain),  concat(gmd:CI_OnlineResource/gmd:linkage/gmd:URL, ' : '))"/>
 
-      <sch:assert test="($contentType = 'Web Service' or $contentType = 'Service Web' or
-              $contentType = 'Dataset' or $contentType = 'Données' or
-              $contentType = 'API' or $contentType = 'Application' or
-              $contentType='Supporting Document' or $contentType = 'Document de soutien')">$locMsgCt</sch:assert>
+      <sch:assert test="$contentType = tokenize($resourceContentTypesListMain, ', ')">$locMsgCt</sch:assert>
```
josegar74: The changes work, but why not do the value check with the thesaurus content instead of the message?

I would update the check to something like this, to unify this kind of thesaurus value checks:

https://github.com/metadata101/iso19139.ca.HNAP/pull/399/files#diff-7c8ff55e3e815d4f6aa802798f973211c9eab945cc11091e8795758a6b60a1deR424-R430

or 

https://github.com/metadata101/iso19139.ca.HNAP/blob/b1fe3249792999af13a0165120906116b2cf736c/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch#L224-L227

tylerjmchugh (approved):
Retested using the following criteria:

- Metadata with English content as the main language and French as the alt language
	- Test with English Interface.
		- Test English Missing and that the error message is English with English list of values. [PASS]
		- Test French Missing and that the error message is English with French list of values. [PASS]
	- Test with French Interface.
		- Test English Missing and that the error message is French with English list of values. [PASS]
		- Test French Missing and that the error message is French with French list of values. [PASS]
- Metadata with English content as the alt language and French as the main language
	- Test with English Interface.
		- Test English Missing and that the error message is English with English list of values. [PASS]
		- Test French Missing and that the error message is English with French list of values. [PASS]
	- Test with French Interface.
		- Test English Missing and that the error message is French with English list of values. [PASS]
		- Test French Missing and that the error message is French with French list of values. [PASS]
		
All tests pass

josegar74 (approved):
Thanks for the updates, tested and works fine.

Issue #395, opened 2024-08-23
Unit of Measurement attribute (uom) in gco:distance 


wangf1122:
The NHAP metadata has such enrichment logic to append http://standards.iso.org/ittf/PubliclyAvailableStandards/ISO_19139_Schemas/resources/uom/gmxUom.xml# as part of the attribute

https://github.com/metadata101/iso19139.ca.HNAP/blob/3f7ed80b30eef0e4ca4f0011f352a60b2e6987de/src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl#L916-L922

There are two issues.

1) The xslt will append the url indefinitely. This can be easily fixed by checking if the url is already part of it

```
  <xsl:template  match="gco:Distance">
    <xsl:element name="gco:{local-name()}">
      <xsl:apply-templates select="@*"/>
      <xsl:choose>
        <xsl:when test="not(starts-with(@uom, 'http://standards.iso.org/ittf/PubliclyAvailableStandards/ISO_19139_Schemas/resources/uom/gmxUom.xml#'))">
          <xsl:attribute name="uom">http://standards.iso.org/ittf/PubliclyAvailableStandards/ISO_19139_Schemas/resources/uom/gmxUom.xml#<xsl:value-of select="@uom"/></xsl:attribute>
        </xsl:when>
        <xsl:otherwise>
          <xsl:attribute name="uom"><xsl:value-of select="@uom"/></xsl:attribute>
        </xsl:otherwise>
      </xsl:choose>
      <xsl:apply-templates select="node()"/>
    </xsl:element>
  </xsl:template>
``` 

2) However issue 1 is not a big concern. But the question is **why such url needs to be added?**. Here is some sample metadata of without such url
![image](https://github.com/user-attachments/assets/b1bf2e60-71c2-4801-b7c9-3c60ab5c4584)

(https://apps.titellus.net/geonetwork/srv/eng/catalog.search#/metadata/77b14f4e-ca84-4786-9bb8-7e59b143fcb7)

But HNAP metadata behave like:
![image](https://github.com/user-attachments/assets/452fa18e-a4bc-46ca-b425-7bd3f7bb7ff3)

So if there is no real benefit of adding such url, we can proceed to remove it and keep it consistent with ISO13139?

fxprunayre:
[...]

PR #397, opened 2024-08-28
fix uom url appending bug. Fix search#/metadata/{UUID} ui issue

wangf1122:
Related to issue https://github.com/metadata101/iso19139.ca.HNAP/issues/395

The url was appended multiple time in the [update-fixed-info.xsl](https://github.com/metadata101/iso19139.ca.HNAP/blob/3f7ed80b30eef0e4ca4f0011f352a60b2e6987de/src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl#L916-L922)

Also the geonetwork/srv/eng/catalog.search#/metadata/{UUID} ui is retrieving the uom from the search engine. Similar code from other schema like:

https://github.com/geonetwork/core-geonetwork/blob/22a87f68d2ec74ddbed86a83e0208fc2a8ce262d/schemas/iso19115-3.2018/src/main/plugin/iso19115-3.2018/index-fields/index.xsl#L701-L703

should apply to HNAP in order to make the display user friendly

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/index-fields/index.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/index-fields/index.xsl
@@ -573,7 +573,9 @@
 
           <xsl:for-each select="gmd:distance/gco:Distance[. != '']">
             <resolutionDistance>
-              <xsl:value-of select="concat(., ' ', @uom)"/>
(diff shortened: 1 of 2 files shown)


josegar74 (approved):
Tested, works fine.

PR #403, opened 2024-10-10
Unable to rename graphicOverview

tylerjmchugh:
Currently when you add a thumbnail and then go back to edit the name the updated name is never saved.

Initial Name:
image
Updated Name:
image
Name After Update:
image

This is caused by the template in `thumbnail-add.xsl` not matching the correct `graphicOverview` element.

Here `position()` is used to match the `resourceIdx` but `position()` does not return the index of the graphic element causing the match to fail.

Also the `resourceHash` is failing to match the generated `md5Hex` value.

This PR aims to fix this issue by changing the template in `thumbnail-add.xsl` to be more like the one in `onlinesrc-add.xsl` which correctly gets the matching position and resource hash.


josegar74 (approved):
It works fine.

Is this change required in core GeoNetwork for iso19139 also?

PR #404, opened 2024-10-17
Spatial representation type mandatory and validation rules

wangf1122:
Spatial Representation type is needed for spatial type metadata. This field is needed as mandatory. So the condition and related validation rule is needed.

Here is the change result.

image

This field can be removed or added by user's decision. So the red asterisk sign wont be shown if it's missing. But it will be showing when it presents.

image

Also the related validation rule is added when this field is showing but leave as empty.

image


tylerjmchugh (approved):
Tested, works as expected:

- Asterisk is shown when the `gmd:spatialRepresentationType` element is added
- Spatial records with no type fails validation
- Spatial records with invalid type fails validation
- Spatial records with valid type pass validation
- Non-spatial records pass validation
- Validation message is shown when the new validation fails

josegar74 (approved):

PR #405, opened 2024-10-17
update linkage field condition

wangf1122:
This is to coordinate with this pull request in Geonetwork to update the required fields.

https://github.com/geonetwork/core-geonetwork/pull/8294

Also this pull request will replace this current pull request https://github.com/metadata101/iso19139.ca.HNAP/pull/391.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/loc/eng/labels.xml
+++ b/src/main/plugin/iso19139.ca.HNAP/loc/eng/labels.xml
@@ -1261,7 +1261,7 @@
         <label>Metadata language</label>
         <btnLabel>Add language</btnLabel>
         <description>Select the language used to create the metadata</description>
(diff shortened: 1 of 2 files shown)


josegar74 (commented):
I have tested the pull request, but for the contacts linkage element the red asterisk is displayed for me:

contact-linkage

wangf1122:
@josegar74 

It was a typo. It should be "optional" as the condition because we made the code change in GN as 
https://github.com/geonetwork/core-geonetwork/blob/bedfe3da2b9461e5549f6c9935a5f9c73a9a4f97/web/src/main/webapp/xslt/ui-metadata/form-builder.xsl#L121

I fixed it and please check again.

josegar74 (approved):
Tested and works fine.

PR #406, opened 2024-10-22
indexing with byte64 data in search engine

wangf1122:
On the search page, catalogue search engine contains some byte data which can speed up the page loading and avoid unwanted backend api querying. HNAP is lack of such indexing mechanism.

image

So this change should be brought from dublin-core

https://github.com/geonetwork/core-geonetwork/blob/cd72f8ec309e866c0fdd2d87ea142dea52627afc/schemas/dublin-core/src/main/plugin/dublin-core/index-fields/index.xsl#L248-L252

Result, 

image

This is related to #[5467](https://github.com/geonetwork/core-geonetwork/pull/5467)

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/index-fields/index.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/index-fields/index.xsl
@@ -391,6 +391,11 @@
             <xsl:if test="normalize-space(../../gmd:fileDescription) != ''">,
               "nameObject": <xsl:value-of select="gn-fn-index:add-multilingual-field('name', ../../gmd:fileDescription, $allLanguages, true())"/>
             </xsl:if>
+            <xsl:variable name="data"
+                          select="util:buildDataUrl(., 140)"/>
+            <xsl:if test="$data != ''">,
+              "data": "<xsl:value-of select="$data"/>"
+            </xsl:if>
             }</overview>
         </xsl:for-each>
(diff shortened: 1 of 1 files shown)


ianwallen:
Also related to the following  #[8466](https://github.com/geonetwork/core-geonetwork/pull/8466)
Which is required before these changes can be applied.

ianwallen (approved):

josegar74 (commented):
It's really strange, the change is in dublin-core, but has been removed later from iso19139:

https://github.com/geonetwork/core-geonetwork/blob/17848bd7b0fcd3cd95f40bca92f2a4506b3ac84a/schemas/iso19139/src/main/plugin/iso19139/index-fields/index.xsl#L376-L384

It seems removed in https://github.com/geonetwork/core-geonetwork/blob/7481a5c025fc5b1b169f8d9591598be8371eb277/schemas/iso19139/src/main/plugin/iso19139/index-fields/index.xsl

In the previous revision, it's there:

https://github.com/geonetwork/core-geonetwork/blob/ebbaa8f1bf5e5018b5eb840a684f3ae5dc31806a/schemas/iso19139/src/main/plugin/iso19139/index-fields/index.xsl

It should be fine the pull request, but I would like to check why it was removed from iso profiles in GeoNetwork.

josegar74 (approved):

PR #408, opened 2024-10-29
Enable show-hints-on-focus for typeahead fields

tylerjmchugh:
Currently fields with typeahead do not give any initial indication that the field has suggested values. This is an issue for new users as the field appears as a plain text field. When a keywordPicker field is validated against its suggested values it is unclear to the user how to see all the options.

This PR aims to fix this issue by introducing the `show-hints-on-focus` feature for typeahead fields. This feature shows all options on the initial click and adds a caret to the field to mock a dropdown list. Adding the caret will make it more clear to the user that they will see some options once they click into the field.

image
image


ianwallen:
Released to https://github.com/geonetwork/core-geonetwork/pull/8401

josegar74 (approved):
Tested, works fine.

PR #409, opened 2024-10-30
Fix http/https inconsistencies with GC_Open_Licenses

tylerjmchugh:
Samples and templates were not updated to use https for `Open Government Licence - Canada` causing them to be invalid with the pre-populated values. 

This PR aims to fix this issue by updating the licenses in samples and templates to use https.

Edit: Split some of the changes into #418

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/inflate-metadata.xsl
@@ -339,16 +339,16 @@
             <gmd:useLimitation xsi:type="gmd:PT_FreeText_PropertyType" gco:nilReason="missing">
               <gco:CharacterString>
                 <xsl:choose>
-                  <xsl:when test="$altLang = 'fra'">Open Government Licence - Canada (http://open.canada.ca/en/open-government-licence-canada)</xsl:when>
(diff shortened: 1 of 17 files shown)


ianwallen (approved):

ianwallen:
Changes look good however it may affect the validation on FGP

@bo-lu, @josegar74 should we be requesting for this to be back ported to 3.6?

bo-lu:
@ianwallen @anthonyfok 

Hi Ian, I am in support of backporting to 3.6.x, but I will have to check with another team to see if there are any dependencies that can cause breaking changes. I seem to recall that the '-' and '−' was an unfortunately error that wasn't caught until another system became dependent on it (via our CSW). 

Regarding HTTP to HTTPS, we will investigate if we can perform a find and replace on the database to upgrade the license to HTTPS and have it default to HTTPS in the future.

Anthony and I are currently prioritizing an OS upgrade on our servers so we will put this in our backlog. It is also due time for us to update the LAC Core Subject Thesaurus.

Btw, our source is now on gccode if you'd like to have a look!

ianwallen:
For this PR - not the backports.  Can you split the PR into 2 parts.  One for the https which we can approve quickly.
The one with the hyphen may take longer to get approval it needs more coordination with FGP

tylerjmchugh:
Split the changes unrelated to https into a new PR: #418

josegar74 (approved):

PR #414, opened 2024-11-13
language country instead of ISO language code

wangf1122:
The language code was defined in the template as : https://github.com/metadata101/iso19139.ca.HNAP/blob/adcfe954b0e7a4b0ac98f043c32a0cd46630c5a0/src/main/plugin/iso19139.ca.HNAP/templates/hnaplevel1-english.xml#L18

However, in the editor mode, the language code is from ISO language code which is different from the original templte. It is adding some confusing to the user.

The purpose of this change is to introduce a new codelist library, and use "eng;CAN" or "fra:CAN" as standard language code. So this library can be expended in the future if new language code as needed.

### DIFF
--- a/src/main/config/codelist/local/thesauri/theme/ISO_Languages_Countries.rdf
+++ b/src/main/config/codelist/local/thesauri/theme/ISO_Languages_Countries.rdf
@@ -28,7 +28,7 @@
             xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">eng; CAN
         </ns2:prefLabel>
     </rdf:Description>
-    <rdf:Description rdf:about="http://geonetwork-opensource.org/EC/languages#FRA">
+    <rdf:Description rdf:about="http://geonetwork-opensource.org/GC/languages#FRA">
             <ns2:scopeNote
                 xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">Definition
             </ns2:scopeNote>
--- a/src/main/config/codelist/local/thesauri/theme/ISO_Languages_Countries.rdf
+++ b/src/main/config/codelist/local/thesauri/theme/ISO_Languages_Countries.rdf
@@ -13,7 +13,7 @@
         <dc:description></dc:description>
       </skos:ConceptScheme>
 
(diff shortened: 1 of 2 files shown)


src/main/config/codelist/local/thesauri/theme/ISO_Languages_Countries.rdf
```diff
@@ -0,0 +1,46 @@
...
+      <skos:ConceptScheme>
+        <dc:title>Languages</dc:title>
+        <dc:title xml:lang="en">Languages</dc:title>
+        <dc:title xml:lang="fr">Langues</dc:title>
+        <dc:publisher xml:lang="en">Government of Canada; Library and Archives Canada</dc:publisher>
+        <dc:publisher xml:lang="fr">Gouvernement du Canada; Bibliothèque et Archives Canada</dc:publisher>
+        <dc:description></dc:description>
+      </skos:ConceptScheme>
+
+    <rdf:Description rdf:about="http://geonetwork-opensource.org/EC/languages#ENG">
```
ianwallen:
```suggestion
    <rdf:Description rdf:about="http://geonetwork-opensource.org/GC/languages#ENG">
```

src/main/config/codelist/local/thesauri/theme/ISO_Languages_Countries.rdf
```diff
@@ -0,0 +1,46 @@
...
+        </ns2:scopeNote>
+        <ns2:prefLabel
+            xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="fr">eng; CAN
+        </ns2:prefLabel>
+        <rdf:type rdf:resource="http://www.w3.org/2004/02/skos/core#Concept"/>
+        <ns2:prefLabel
+            xmlns:ns2="http://www.w3.org/2004/02/skos/core#" xml:lang="en">eng; CAN
+        </ns2:prefLabel>
+    </rdf:Description>
+    <rdf:Description rdf:about="http://geonetwork-opensource.org/EC/languages#FRA">
```
ianwallen:
```suggestion
    <rdf:Description rdf:about="http://geonetwork-opensource.org/GC/languages#FRA">
```

josegar74 (approved):
Tested, works fine.

PR #416, opened 2024-11-27
resource name tool tip

wangf1122:
The tooltip of resource name is missing

image

It is defined in https://github.com/metadata101/iso19139.ca.HNAP/blob/adcfe954b0e7a4b0ac98f043c32a0cd46630c5a0/src/main/plugin/iso19139.ca.HNAP/config/associated-panel/default.json#L25

So the tooltip text is moved from:
https://github.com/metadata101/iso19139.ca.HNAP/blob/adcfe954b0e7a4b0ac98f043c32a0cd46630c5a0/src/main/plugin/iso19139.ca.HNAP/loc/eng/labels.xml#L1464

This pull request change also update gmd:version and gmd:description to have better help text.


ianwallen (approved):

josegar74 (approved):
Tested , works fine.

Issue #415, opened 2024-11-25
Unit of Meansurement (uom) rendering issue


wangf1122:
The current UI is not able to handle HNAP uom spec

![image](https://github.com/user-attachments/assets/4bdd8601-81f1-4c22-b61d-eb6aab1a741b)

~~~
<gco:Distance uom="http://standards.iso.org/ittf/PubliclyAvailableStandards/ISO_19139_Schemas/resources/uom/gmxUom.xml#m">1</gco:Distance>
~~~

The ISO13139 standard has no such url as part of the uom part.

~~~
 <gco:Distance uom="m">10</gco:Distance>
~~~

So wonder why HNAP needs such url and how to make the UI handle such url?

josegar74:
I've seen both cases. The file `gmxUom.xml` seems containing kind of codelist https://github.com/geonetwork/core-geonetwork/blob/main/schemas/iso19139/src/main/plugin/iso19139/schema/resources/uom/gmxUom.xml

I don't think HNAP enforces it, at least I don't find that in the specification.

In any case, the code enforcing it is https://github.com/metadata101/iso19139.ca.HNAP/blob/adcfe954b0e7a4b0ac98f043c32a0cd46630c5a0/src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl#L915-L929

Maybe can be added a variable in the xsl to enforce the URL or use the provided value as it is.

wangf1122:
@josegar74 

Do you mean in the current HNAP schema enforcing logic which is the [/update-fixed-info.xsl](https://github.com/metadata101/iso19139.ca.HNAP/blob/adcfe954b0e7a4b0ac98f043c32a0cd46630c5a0/src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl#L915-L929)

We should not append the url to the unit. We should use the unit as is? What I understand it the codelist in HNAP which been pointed is same as what ISO19139 looks like. So I dont see any advantage to add such url in front of the unit like:

"http://standards.iso.org/ittf/PubliclyAvailableStandards/ISO_19139_Schemas/resources/uom/gmxUom.xml#m"
instead of
"m"


I have created a pull request to get this url removed in update-fixed-info

https://github.com/metadata101/iso19139.ca.HNAP/pull/417 and make it same as how indexing works
[...]

PR #417, opened 2024-11-28
remove url as part of unit of measurement

wangf1122:
The url appended to the uom attribute (unit of measurement) is causing some issue https://github.com/metadata101/iso19139.ca.HNAP/issues/415

The discussion is to not enforce such url which is brought from indexing

https://github.com/metadata101/iso19139.ca.HNAP/blob/adcfe954b0e7a4b0ac98f043c32a0cd46630c5a0/src/main/plugin/iso19139.ca.HNAP/index-fields/index.xsl#L576-L578

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/update-fixed-info.xsl
@@ -912,24 +912,6 @@
   </xsl:template>
 
 
-  <xsl:template  match="gco:Distance">
-    <xsl:element name="gco:{local-name()}">
-      <xsl:apply-templates select="@*"/>
-      <xsl:choose>
-        <!--Avoid append the url recursively. Only append the url once. -->
-        <xsl:when test="starts-with(@uom, 'http://standards.iso.org/ittf/PubliclyAvailableStandards/ISO_19139_Schemas/resources/uom/gmxUom.xml#')">
-          <xsl:attribute name="uom"><xsl:value-of select="concat(., ' ', tokenize(@uom, '#')[2])"/></xsl:attribute>
-        </xsl:when>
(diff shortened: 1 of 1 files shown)


josegar74 (approved):
@wangf1122 An alternative would be to remove the template to match `gco:Distance` from `update-fixed-info`, that way it preserves what the user enters. This is how it works in https://github.com/geonetwork/core-geonetwork/blob/4.2.x/schemas/iso19139/src/main/plugin/iso19139/update-fixed-info.xsl

But in any case the solution looks fine.

wangf1122:
> @wangf1122 An alternative would be to remove the template to match `gco:Distance` from `update-fixed-info`, that way it preserves what the user enters. This is how it works in https://github.com/geonetwork/core-geonetwork/blob/4.2.x/schemas/iso19139/src/main/plugin/iso19139/update-fixed-info.xsl
> 
> But in any case the solution looks fine.

@josegar74 

Removing this template from update-fixed-info.xl seems to be the right choice and this template should never been there. My only concern is what if there are production data already come with this url which the harm is already done. We will need to do some manual data fixing. If you are ok with this, then I can just get rid of this unnecessary template

josegar74:
@wangf1122  in any case the update must be done for the existing metadata, if we keep the template with the value change, the existing metadata will require editing for the update-fixed-info to be applied.

I think we can simply delete the template and apply a database update to remove the unwanted URL in the attribute.

wangf1122:
@josegar74 

I have removed the whole xsl template. So it will not add such url and will do nothing to gco:distance/@uom attribute. As discussed, we could deal with individual data by manual editing or database direct update.

josegar74:
@wangf1122 thanks, I have retested the latest changes and they look good.

PR #418, opened 2024-12-11
Fix inconsistencies with hyphens for GC_Open_Licenses

tylerjmchugh:
Some licenses use en dash (`–`) instead of hyphen (`-`) causing issues with typeahead as not all options are shown when one of the two dashes are typed into the field.

This PR aims to fix this issue by making all the licenses consistently use hyphen instead of en dash.

Additionally there is a typo in the French Newfoundland and Labrador license.

Split from #409


ianwallen:
@bo-lu 
Any updates related to your comment on PR #409 
https://github.com/metadata101/iso19139.ca.HNAP/pull/409#issuecomment-2450673248

PR #424, opened 2025-02-19
Removed subtuitionGroup for some gmx schema. Add labels

wangf1122:
In the advanced view, there are a few labels missing for the data quality section. Also some substitutional schema popped out. For example, [gmd:level](https://www.datypic.com/sc/niem21/e-gmd_level-1.html) only have [gmd:MD_ScopeCode](https://www.datypic.com/sc/niem21/e-gmd_MD_ScopeCode.html)

But gmx:MX_ScopeCode is also showing.
image

So of the xsd change to get the substitutionGroup removed. Also added couple of missing labels for

image
and
image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schema/gmx/crsItem.xsd
+++ b/src/main/plugin/iso19139.ca.HNAP/schema/gmx/crsItem.xsd
@@ -113,7 +113,7 @@
 		</xs:complexContent>
 	</xs:complexType>
 	<!-- ........................................................................ -->
-	<xs:element name="ML_CompoundCRS" type="gmx:ML_CompoundCRS_Type" substitutionGroup="gml:CompoundCRS"/>
+	<xs:element name="ML_CompoundCRS" type="gmx:ML_CompoundCRS_Type"/>
 	<!-- ........................................................................ -->
 	<xs:complexType name="ML_CompoundCRS_PropertyType">
 		<xs:sequence minOccurs="0">
@@ -175,7 +175,7 @@
 		</xs:complexContent>
 	</xs:complexType>
 	<!-- ........................................................................ -->
-	<xs:element name="ML_GeodeticCRS" type="gmx:ML_GeodeticCRS_Type" substitutionGroup="gml:GeodeticCRS"/>
+	<xs:element name="ML_GeodeticCRS" type="gmx:ML_GeodeticCRS_Type"/>
 	<!-- ........................................................................ -->
 	<xs:complexType name="ML_GeodeticCRS_PropertyType">
 		<xs:sequence minOccurs="0">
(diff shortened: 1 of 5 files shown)


src/main/plugin/iso19139.ca.HNAP/schema/gmx/crsItem.xsd
```diff
@@ -113,7 +113,7 @@
 		</xs:complexContent>
 	</xs:complexType>
 	<!-- ........................................................................ -->
-	<xs:element name="ML_CompoundCRS" type="gmx:ML_CompoundCRS_Type" substitutionGroup="gml:CompoundCRS"/>
+	<xs:element name="ML_CompoundCRS" type="gmx:ML_CompoundCRS_Type"/>
```
ianwallen, line 116: The schema cannot be changed.  It is part of a standard and the only way the schema can be changes is if the standard is changed.  So generally, no files in the schema folder can be changed unless there is an obvious issue with the implementation of the standard.

The source for the schema file is here
https://schemas.metadata.geo.ca/metadata/tools/schemas/metadata/can-cgsb-171.100-2009-a/gmx/crsItem.xsd
And older reference here
https://nap.geogratis.gc.ca/metadata/tools/schemas/metadata/can-cgsb-171.100-2009-a/gmx/crsItem.xsd
And it is based on the iso19139 2007
https://github.com/geonetwork/core-geonetwork/blob/main/schemas/iso19139/src/main/plugin/iso19139/schema/gmx/crsItem.xsd

wangf1122: if so, for example, the gml:baseUnit and gmx:baseUnit as I can see are like duplicated label and they look like to be meant for the same thing. I guess it's ok for this scenario?

ianwallen: When I mentioned `obvious issues` above, I did not mean obvious issues with the standard.  I meant obvious issues with the implementation of the standard in the HNAP implementation.
You seem to be raising issues with the standard and we cannot change that.

I don't believe there are exceptions to changing the schema.  Otherwise anyone who gets an NAP metadata record will not be able to properly validate the schema against the standard NAP schema.  Sorry but we need to follow the standard schema as it is documented - even if it has issues.

josegar74:
@wangf1122 maybe try https://github.com/metadata101/iso19139.ca.HNAP/blob/4.4.x/src/main/plugin/iso19139.ca.HNAP/schema-substitutes.xml instead of changing the xsd.

I have never used it, but it seems intended to limit the substitutes offered by the metadata editor.

wangf1122:
@josegar74 

I had looked this file. It seems can only solve the issue of "gmd:MD_ScopeCode". But others.

You can see I placed
image


But the other name is still showing.

image

josegar74:
@wangf1122 I tried the following:

```xml
 <field name="gml:UnitDefinition">
    <substitute name="gml:BaseUnit"/>
    <substitute name="gml:ConventionalUnit"/>
    <substitute name="gml:DerivedUnit"/>
    <substitute name="gml:UnitDefinition" />
  </field>
```

but it doesn't show up the `gml:UnitDefinition` entry:

substitutions

wangf1122:
@josegar74  @ianwallen 

I made a separated pull request (https://github.com/metadata101/iso19139.ca.HNAP/pull/447) just to address the md_scope issue. Maybe we can get that change in first?

PR #431, opened 2025-03-18
Fix unwanted date time if for resourceDate index

wangf1122:
The XML of the record contains no such datetime or date information but just simple year.

image

However the record view page added extra information

image

In the xsl-view mode which was sent by notification email (for example:
https://YOUR_HOST/catalogue/srv/eng/catalog.search#/metadata/36f1a5bf-9a28-4c05-9acc-f838de67b167/formatters/xsl-view?root=div&view=advanced). The view renders as is.

image

It created confusion to the user. If the original xml contains no such date, datetime, or zone information, we should not add extra information to confuse the user. Therefore, this pull request is to remove the conversion from the index for this specific "resourceDate" field


josegar74 (approved):
Tested, works fine.

PR #436, opened 2025-04-01
translate fra default language to fre in indexing

wangf1122:
The indexing for several field are marked as langfra which is causing some potential issue because Geonetwork identify "fre" as the three letters French language code not "fra"

image

This is mainly causing by parsing the variable from

https://github.com/metadata101/iso19139.ca.HNAP/blob/c1b89e308d1edff91dbda89108fdc81e180219e3/src/main/plugin/iso19139.ca.HNAP/templates/hnaplevel1-french.xml#L18

So this pull request will do some pre-transformation based on that fra code to fre.


ianwallen (approved):

josegar74 (approved):
Tested, it works fine.

josegar74 (commented):
Tested, works fine.

PR #440, opened 2025-04-03
contactForResource is incorrectly indexed

tylerjmchugh:
Currently the `contactForResource` index field is never set as it uses `gmd:identificationInfo/gmd:MD_DataIdentification/gmd:pointOfContact` instead of `gmd:identificationInfo/gmd:MD_DataIdentification/gmd:citation/gmd:CI_Citation/gmd:citedResponsibleParty`.

This causes issues such as a header without any content in the record view:
image

Another issue is that the contact is not shown as an option in the feedback form.

This PR aims to fix this issue by setting the `contactForResource` to `gmd:identificationInfo/gmd:MD_DataIdentification/gmd:citation/gmd:CI_Citation/gmd:citedResponsibleParty`.

The following help text suggests that these two should be mapped together:
image

After these changes the record view displays like:
image

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/index-fields/index.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/index-fields/index.xsl
@@ -372,6 +372,13 @@
           <xsl:for-each select="gmd:edition/*">
             <xsl:copy-of select="gn-fn-index:add-field('resourceEdition', .)"/>
           </xsl:for-each>
+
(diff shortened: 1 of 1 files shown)


josegar74 (approved):
I haven't test it, but it looks fine as HNAP uses `gmd:citedResponsibleParty` instead of `gmd:pointOfContact` for the resource contact information.

PR #443, opened 2025-04-16
Index resource date fields as defined in the metadata.

josegar74:
Analog change to https://github.com/geonetwork/core-geonetwork/pull/8365

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/index-fields/index.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/index-fields/index.xsl
@@ -296,13 +296,13 @@
             <xsl:choose>
               <xsl:when test="$zuluDateTime != ''">
                 <xsl:element name="{$dateType}DateForResource">
-                  <xsl:value-of select="$zuluDateTime"/>
+                  <xsl:value-of select="$date"/>
                 </xsl:element>
                 <xsl:element name="{$dateType}YearForResource">
-                  <xsl:value-of select="substring($zuluDateTime, 0, 5)"/>
+                  <xsl:value-of select="substring($date, 0, 5)"/>
                 </xsl:element>
                 <xsl:element name="{$dateType}MonthForResource">
-                  <xsl:value-of select="substring($zuluDateTime, 0, 8)"/>
+                  <xsl:value-of select="substring($date, 0, 8)"/>
                 </xsl:element>
               </xsl:when>
               <xsl:otherwise>


tylerjmchugh (approved):
Tested, works fine

src/main/plugin/iso19139.ca.HNAP/index-fields/index.xsl
```diff
@@ -296,13 +296,13 @@
             <xsl:choose>
               <xsl:when test="$zuluDateTime != ''">
```
ianwallen, line 297: Why are we checking for 
$zuluDateTime != ''"
But then not using it?

Also the issue reported seems to be to not assigning zulu dates when there are no timezones specified in the date.  This makes sense however why are we not performing a check to find out if there is a timezone in the date?

```
<xsl:if test="matches($date, '(\+|-)\d{2}:\d{2}|Z$')">
  <!-- Has timezone -->
</xsl:if>
```
- (\+|-)\d{2}:\d{2} matches a timezone like +02:00 or -05:00
- Z$ matches a Z at the end (Zulu time / UTC)

i.e. I would expect 

```
      <xsl:choose>
         <xsl:when test="matches($date, '(\+|-)\d{2}:\d{2}|Z$')''">
               <xsl:element name="{$dateType}DateForResource">
                  <xsl:value-of select="$zuluDateTime"/>
                </xsl:element>
                <xsl:element name="{$dateType}YearForResource">
                  <xsl:value-of select="substring($zuluDateTime, 0, 5)"/>
                </xsl:element>
                <xsl:element name="{$dateType}MonthForResource">
                  <xsl:value-of select="substring($zuluDateTime, 0, 8)"/>
                </xsl:element>
          </xsl:when>
         <xsl:when test="$date != ''">
               <xsl:element name="{$dateType}DateForResource">
                  <xsl:value-of select="$date"/>
                </xsl:element>
                <xsl:element name="{$dateType}YearForResource">
                  <xsl:value-of select="substring($date, 0, 5)"/>
                </xsl:element>
                <xsl:element name="{$dateType}MonthForResource">
                  <xsl:value-of select="substring($date, 0, 8)"/>
                </xsl:element>
           </xsl:when>
           <xsl:otherwise>
                <indexingErrorMsg type="object">
                  {
                    "string": "indexingErrorMsg-invalidDateFormat",
                    "type": "warning",
                    "values": {
                      "dateType": "<xsl:value-of select="util:escapeForJson($dateType)"/>",
                      "date": "<xsl:value-of select="util:escapeForJson($date)"/>"
                    }
                  }
                </indexingErrorMsg>
              </xsl:otherwise>
            </xsl:choose>
```

josegar74: @ianwallen 

> Why are we checking for
$zuluDateTime != ''"
But then not using it?

Because if date is not valid `$zuluDateTime` contains an empty string. It's used only for that, and then is stored the original metadata date.

ianwallen: OK - thank you

ianwallen (approved):

jodygarnett:
@ianwallen / @josegar74 is this PR needed?

PR #441, opened 2025-04-11
Fix metadata language display in the metadata editor

josegar74:
Without the fix the metadata language in the metadata editor displays a dropdown to select a thesaurus:

Screenshot 2025-04-11 at 16 10 27

Related to https://github.com/metadata101/iso19139.ca.HNAP/pull/414

Issue #145, opened 2020-12-02
Changing the main language


ianwallen:
As HNAP requires the main language the main language was added to the Simple editor form as read only.  But it remained on the full view as editable.

It was done that way because changing the main language of the document without changing all the locales messes up the metadata record and confuses the editor.  But it should work as expected for unilingual metadata - which I believe is used by others. So maybe making the field read-only was not a great idea.

I think we should look into seeing if there is a better way of applying the language change.  Then we could allow users to change the language.

I'm not sure if this is possible in the editor?  
Looking for some feedback on this issue.

fxprunayre:
This is a discussion we had on couple of projects here and we never moved forward on this but the idea was to move the language out of the main editor form and add a specific component to add/remove and change default language. The component can then call specific API/XSL process to take care of the language change logic.

PR #444, opened 2025-04-17
Fix metadata language shows thesaurus dropdown in basic view

tylerjmchugh:
Currently the basic editor view displays the metadata language with a thesaurus selector.
image

The thesaurus selector should not be shown as this field should only allow values from `ISO_Languages_Countries`. Additionally this field is read-only so the thesaurus selector does nothing.

The thesaurus selector is shown because the `data-thesaurus-key="external.theme.ISO_Languages_Countries"` directive attribute is not passed to the input in basic/default/flat mode.

This PR aims to fix this issue by updating the custom template to pass the directive attributes which contains the thesaurus key.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
+++ b/src/main/plugin/iso19139.ca.HNAP/layout/layout-custom-fields.xsl
@@ -816,6 +816,9 @@
       <xsl:with-param name="cls" select="local-name()"/>
       <xsl:with-param name="xpath" select="$xpath"/>
       <xsl:with-param name="type" select="gn-fn-metadata:getFieldType($editorConfig, name(), '', $xpath)"/>
+      <xsl:with-param name="directiveAttributes">
+        <xsl:copy-of select="gn-fn-metadata:getFieldDirective($editorConfig, name(),name(*[@codeListValue]), $xpath)"/>
+      </xsl:with-param>
       <xsl:with-param name="name" select="''"/>
       <xsl:with-param name="editInfo" select="*/gn:element"/>
       <xsl:with-param name="parentEditInfo" select="gn:element"/>


josegar74:
@tylerjmchugh this is a duplicated of https://github.com/metadata101/iso19139.ca.HNAP/pull/441

tylerjmchugh:
@josegar74 There is an issue with the advanced view in #441. This PR does not have the issue. If the advanced view issue is to be resolved in #441 I will abandon this PR.

See #441

josegar74 (approved):
@tylerjmchugh it works fine, we can use this one instead of https://github.com/metadata101/iso19139.ca.HNAP/pull/441. But in the advance view the field is editable. I see this commit from @ianwallen adding that behaviour:

https://github.com/metadata101/iso19139.ca.HNAP/commit/e37fe74a83d879454378eebbfbc7224914ce65a4

I'm not really sure why in the basic view can cause issues and not in the advanced view, but other than that the pull request works.

ianwallen:
@josegar74 - yes the language is editable in the advanced view as users want to be able to change the main language.  

The problem is that there is bug where if they change the main language, it would then need to update all the xml to move the <gco:CharacterString> to <gmd:LocalisedCharacterString> and vice versa.  There was a discussion on this issue https://github.com/metadata101/iso19139.ca.HNAP/issues/145 but it was never actioned.

josegar74:
But if users change the language from the advance view, the issue does not happen?

tylerjmchugh:
> But if users change the language from the advance view, the issue does not happen?

@josegar74 I just tested and the issue does still occur. I think the reason it is disabled in basic view is so the language is not changed accidentally. With it enabled in advanced view it still gives the user the ability to change the language and manually fix any errors if they really need to.

josegar74:
I see, another option can be to use different templates for main language or modify the `update-fixed-info.xsl` to track language change, but that is more complex and out of the scope of this pull request.

PR #458, opened 2025-10-21
Check for Open Government license specifically

tylerjmchugh:
Currently if we want to add additional licenses we need to update the `GC_Open_Licenses.rdf`. This however introduces an issue as that new license would also cause the record to pass validation.

This PR aims to solve this issue by checking for Open Government licenses specifically instead of checking for any license in the `GC_Open_Licenses.rdf`.

With these changes we can now add additional optional licenses to `GC_Open_Licenses.rdf` to be displayed as suggestions while still requiring at least one Open Government license.

Additionally the required licenses are set using a `required-licenses` variable at the top of the schematron to facilitate customization.


jodygarnett:
So the challenge is then that GC_Open_Licenses.rdf ... contains licences that are not Open Government Licenses. Okay I think I can see how that makes sense.

jodygarnett (approved):

PR #462, opened 2025-12-02
Duplicate name tile for contains operations on HNAP services

juanluisrp:
Fix the Parameter type name and value type name labels and helper lists in services metadata.

The context must be an absolute XPath if an XPath is used.

image


juanluisrp:
After the change the labels are fixed and a helper list with recommended values is shown.

image

ianwallen:
Tested locally and works

image

ianwallen (approved):

PR #471, opened 2026-03-04
Skip ZuluDateTime conversion when no timezone is set

PascalLike:
Do not use convertToISOZuluDateTime when no timezone is indicated in the date.


src/main/plugin/iso19139.ca.HNAP/index-fields/index.xsl
tylerjmchugh, line 358: I think the same needs to be done for this date too. This is where the UI gets the publication and creation date from.

PascalLike: Good catch, thank you @tylerjmchugh

tylerjmchugh (approved):
LGTM

juanluisrp (approved):
LGTM.

PR #472, opened 2026-03-19
Unify validations for organisation names for the different types of contacts

josegar74:
Validation errors were duplicated for Cited responsible party and Distributor contacts and different from Metadata contacts:

contact-validation-1

Metadata Contacts:

contact-validation-2

Also non-multilingual validations are done with this field, which is multilingual.

This change:

- Unifies the multilingual validations for the different contacts.
- Removes the non-multilingual validations for organisation name.
- Adds comments and changes some variable names to make them clear.
- Fixes alternative language checks.

### DIFF
--- a/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
+++ b/src/main/plugin/iso19139.ca.HNAP/schematron/schematron-rules-multilingual.sch
@@ -244,40 +244,46 @@
     </sch:rule>
 
 
-
-
     <!-- Contact - Organisation Name -->
+    <!-- Organisation format: Title;Department/Agency;Sub-organizations -->
     <sch:rule context="//gmd:contact/*/gmd:organisationName">
-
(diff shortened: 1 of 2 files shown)


ianwallen:
@josegar74 - my understanding is that you only need to enable the multilingual schematron or the non multilingual validation but don't need both.

In our system, we have disabled the non-multilingual so we don't see those duplicated error.  I believe there are lots of other cases where there are duplicated errors if running both schematrons.

Maybe we should investigate disabling the non-multilingual schematron as a default during initial setup?  May want to get @jodygarnett input on this as I believe he may have had a client in the past that required non multilingual (but not sure?)

PascalLike:
It's correct that the two schematrons shouldn't be activated simultaneously for each metadata item. The changes to the multilingual schematron are still interesting improvements; perhaps the changes to the non-multilingual schematron should be ignored.

josegar74:
@ianwallen this can be accomplished configuring the non-multilingual validation with the XPATH to check for the locale.

Is it ok for you what @PascalLike proposes?

josegar74:
@ianwallen, do you have any comments on my previous message so we can properly update the pull request and merge it?

